PR #313, opened 2026-04-10
chore(release): bump to 0.4.0 and derive version from metadata

Umaaz:
## Bump to 0.4.0 and dynamic Python agent versioning
✨ **Improvement** · ♻️ **Refactor** · 🔧 **Chore**

Bumps the project version to 0.4.0 and implements dynamic version retrieval for the Python agent. This refactors the release pipeline to be more robust and removes redundant source-code modifications during CI.

### Complexity
🟡 Moderate · `4 files changed, 14 insertions(+), 8 deletions(-)`

This PR modifies foundational versioning logic and release infrastructure across both Python and UI components. While the code changes are small, they affect the build/release pipeline and require synchronization between package manifests, source code, and CI scripts.

### Tests
🧪 Existing version tests in the agent are currently out of sync with the new version and dynamic lookup logic.

### Review focus

- **`importlib.metadata` fallback** — consider adding a try/except block for cases where the package is run in an uninstalled development environment.
- **Workflow synchronization** — the `publish-preview.yml` and `publish-dev.yml` workflows still contain redundant `sed` commands for `__init__.py` that should be removed.
- **UI lockfile mismatch** — verify and fix the regression in `ui/package-lock.json` which is currently out of sync with `main`.
- **Broken version test** — `agent/tests/opentrace_agent/test_init.py` still asserts a version of `0.1.0` and will fail in CI.

---

<details>

### Dynamic Versioning
The Python agent now leverages `importlib.metadata` to read its version directly from the package installation. This eliminates the need for the release workflow to perform string replacements on source files, reducing the risk of merge conflicts and build inconsistencies.

### Idempotent Releases
The release workflow has been updated to check for existing GitHub releases before attempting creation. If a release already exists (e.g., created via the UI or a previous partial run), the workflow will now upload assets and update notes instead of failing.

### Note on Branch State
This branch diverged prior to the 0.3.0 release. As a result, `ui/package-lock.json` has regressed to version `0.1.1` in this PR (compared to `0.3.0` on `main`). A rebase or manual synchronization of the lockfile and `CHANGELOG.md` is required before merging.

</details>

### DIFF
--- a/agent/pyproject.toml
+++ b/agent/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "opentraceai"
-version = "0.3.0"
+version = "0.4.0"
 description = "OpenTrace Python agent — maps system architecture, code structure, and service relationships into a knowledge graph"
 readme = "README.md"
 license = "Apache-2.0"
--- a/agent/src/opentrace_agent/__init__.py
+++ b/agent/src/opentrace_agent/__init__.py
@@ -14,4 +14,9 @@
 
 """OpenTrace Python Agent."""
 
-__version__ = "0.3.0"
+from importlib.metadata import PackageNotFoundError, version
+
+try:
+    __version__ = version("opentraceai")
+except PackageNotFoundError:
+    __version__ = "unknown"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,6 @@ jobs:
 
           # Bump agent version
           sed -i "s/^version = .*/version = \"${VERSION}\"/" agent/pyproject.toml
-          sed -i "s/^__version__ = .*/__version__ = \"${VERSION}\"/" agent/src/opentrace_agent/__init__.py
 
           # Bump UI version
           cd ui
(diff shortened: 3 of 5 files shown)


agent/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "opentraceai"
```
ghost, line 2: ⚠️ The version bump here will cause `agent/tests/opentrace_agent/test_init.py` to fail because it asserts `__version__ == "0.1.0"`. That test should be updated or made dynamic.

agent/src/opentrace_agent/__init__.py
```diff
@@ -14,4 +14,6 @@
 
 """OpenTrace Python Agent."""
 
-__version__ = "0.1.0"
+from importlib.metadata import version
+
+__version__ = version("opentraceai")
```
ghost: ⚠️ This will raise `PackageNotFoundError` if the package is not installed (e.g., when running directly from source via `PYTHONPATH`). Consider adding a fallback.

```suggestion
try:
    from importlib.metadata import version, PackageNotFoundError
except ImportError:
    # For Python < 3.8
    from importlib_metadata import version, PackageNotFoundError

try:
    __version__ = version("opentraceai")
except PackageNotFoundError:
    __version__ = "unknown"

```

agent/pyproject.toml
Umaaz: Fixed in 696a26c6 — updated test_init.py to assert dynamically (string type + non-empty) instead of a hardcoded version.

agent/src/opentrace_agent/__init__.py
Umaaz: Fixed in 696a26c6 — added PackageNotFoundError fallback (falls back to "unknown"). Skipped the Python < 3.8 guard since pyproject.toml requires >=3.12.

PR #323, opened 2026-04-13
fix(agent): allow concurrent index while MCP server is running

Umaaz:
## Allow concurrent indexing while MCP server is running
✨ **Improvement** · ♻️ **Refactor** · 🧪 **Tests**

Allows `opentrace index` to run concurrently while an MCP server is active by using a staging-file + atomic-swap pattern. 

The MCP server now monitors the database inode and transparently reopens the file when it detects an update.

### Complexity
🟡 Moderate · `4 files changed, 323 insertions(+), 25 deletions(-)`

This PR introduces a new state management pattern across the CLI and MCP server to handle file-system-level atomicity. It requires careful review of resource lifecycles (closing/reopening connections) and the use of closures for dynamic dependency injection in the MCP server.

### Tests
🧪 Includes unit tests for the reload logic and integration tests verifying the MCP server picks up new data after a file swap.

### Review focus

- **Resource Lifecycle** — verify that `_ReloadableStore` correctly closes old connections before reopening to prevent file descriptor leaks.
- **Atomic Replace** — check the `os.replace` logic in the `index` command to ensure the staging file is cleaned up on failure.

---

<details>

### The Problem
LadybugDB uses exclusive process-level file locking. Even opening a database with `read_only=True` acquires an exclusive lock, meaning a long-running MCP server process completely blocks the `opentrace index` command from updating the knowledge graph.

### The Solution: Staging and Atomic Swaps
1.  **Indexing**: The `index` command now writes to `index.db.staging`. Once the index is complete, it calls `os.replace(staging, live)`, which is atomic on POSIX (and generally on Windows for non-open files).
2.  **Detection**: The MCP server cannot hold the file open indefinitely if we want to replace it. However, keeping it open is necessary for performance. The new `_ReloadableStore` wrapper checks the file's inode (`st_ino`) before each tool execution.
3.  **Reloading**: If the inode changes (indicating the file was replaced by a new index run), the wrapper closes the old connection and opens a new one to the updated file.

### Architectural Changes
- `create_mcp_server` was updated to accept a `Callable` store provider. This allows the FastMCP tools to resolve the "current" active store instance at runtime rather than being bound to a stale object.
- The `GraphStoreAdapter` and `GraphStore` lifecycles are explicitly managed during the swap to prevent connection leaks.

</details>

### DIFF
--- a/agent/src/opentrace_agent/cli/main.py
+++ b/agent/src/opentrace_agent/cli/main.py
@@ -382,14 +382,15 @@ def stats(db_path: str | None, output_format: str) -> None:
 
 
 class _ReloadableStore:
-    """Mutable GraphStore handle that reopens when the DB file is replaced.
+    """Transparent proxy for GraphStore that reopens when the DB file is replaced.
 
     LadybugDB uses exclusive file locking, so a long-running reader (MCP)
     blocks writers (``opentrace index``).  The indexer works around this by
     writing to a staging file and atomically renaming it over the original.
 
-    This class detects the rename (via inode change) and transparently
-    reopens the database on the next access.
+    This proxy detects the rename (via inode change) and transparently
+    reopens the database.  Attribute access is delegated to the inner
+    store, so callers treat this as a regular ``GraphStore``.
     """
 
     def __init__(self, db_path: str | None, store: object | None) -> None:
@@ -399,6 +400,8 @@ class _ReloadableStore:
         self._lock = threading.Lock()
         self._log = logging.getLogger("opentrace_agent.mcp.reload")
 
+    # -- inode helpers -------------------------------------------------------
+
     def _stat_inode(self) -> int | None:
         if self._db_path is None:
             return None
@@ -407,20 +410,18 @@ class _ReloadableStore:
         except OSError:
             return None
 
-    def get(self) -> object | None:
-        """Return the current store, reopening if the DB file changed.
+    def _maybe_reload(self) -> None:
+        """Reopen the database if the file's inode has changed.
 
-        Thread-safe: FastMCP may dispatch tool calls from a thread pool,
-        so we serialize the inode check + reopen with a lock.
+        Thread-safe: FastMCP may dispatch tool calls from a thread pool.
         """
         if self._db_path is None:
-            return self._store
+            return
 
         with self._lock:
-            # Re-check inode under the lock (double-checked).
             current_inode = self._stat_inode()
             if current_inode == self._inode:
-                return self._store
+                return
 
             self._log.info(
                 "Database file changed (inode %s → %s), reopening",
@@ -449,7 +450,18 @@ class _ReloadableStore:
                 except Exception as e:
                     self._log.warning("Failed to reopen database: %s", e)
 
-            return self._store
+    # -- proxy protocol ------------------------------------------------------
+
+    def __bool__(self) -> bool:
+        self._maybe_reload()
+        return self._store is not None
+
+    def __getattr__(self, name: str) -> object:
+        self._maybe_reload()
+        store = self._store
+        if store is None:
+            raise AttributeError(name)
+        return getattr(store, name)
 
     def close(self) -> None:
         with self._lock:
@@ -511,7 +523,7 @@ def mcp_cmd(db_path: str | None, verbose: bool) -> None:
     signal.signal(signal.SIGTERM, _shutdown)
 
     try:
-        server = create_mcp_server(reloadable.get)
+        server = create_mcp_server(reloadable)
         log.debug("MCP server running on stdio")
         server.run(transport="stdio")
     except KeyboardInterrupt:
--- a/agent/src/opentrace_agent/cli/main.py
+++ b/agent/src/opentrace_agent/cli/main.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import logging
+import os
 import signal
 import time
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


agent/src/opentrace_agent/cli/main.py
```diff
@@ -184,28 +185,42 @@ def index(
...
+    except BaseException:
+        # Clean up the staging file on failure.
+        Path(staging_db).unlink(missing_ok=True)
+        raise
 
-    store.close()
+    # Atomically replace the live database.  Readers holding the old
+    # file descriptor continue to see the previous data until they
+    # detect the inode change and reopen.
+    os.replace(staging_db, resolved_db)
```
ghost, line 226: ⚠️ `os.replace` will fail with a `PermissionError` on Windows if `resolved_db` is currently open by another process (like the MCP server). This means concurrent indexing might still not work on Windows.

If Windows support is important, you might need to handle this error or use a different strategy for Windows. For Linux/macOS, this works as intended.

agent/src/opentrace_agent/cli/main.py
```diff
@@ -363,6 +378,67 @@ def stats(db_path: str | None, output_format: str) -> None:
...
+                self._store.close()
+            except Exception:
+                pass
+
+        self._store = None
+        self._inode = current_inode
+
+        if current_inode is not None:
+            from opentrace_agent.store import GraphStore
+
```
ghost, line 447: ⚠️ This is not thread-safe. Since `FastMCP` tools may run in parallel (in a thread pool), multiple threads might detect an inode change simultaneously, leading to multiple reloads, race conditions on `self._store`, and potential resource leaks. 

I recommend adding a `threading.Lock` to protect this method.

```suggestion
    def __init__(self, db_path: str | None, store: object | None) -> None:
        import threading

        self._db_path = db_path
        self._store = store
        self._inode = self._stat_inode()
        self._lock = threading.Lock()
        self._log = logging.getLogger("opentrace_agent.mcp.reload")

    def _stat_inode(self) -> int | None:
        if self._db_path is None:
            return None
        try:
            return os.stat(self._db_path).st_ino
        except OSError:
            return None

    def get(self) -> object | None:
        """Return the current store, reopening if the DB file changed."""
        if self._db_path is None:
            return self._store

        with self._lock:
            current_inode = self._stat_inode()
            if current_inode == self._inode:
                return self._store

            # File was replaced (or removed).
            self._log.info("Database file changed (inode %s → %s), reopening", self._inode, current_inode)

            new_store = None
            if current_inode is not None:
                from opentrace_agent.store import GraphStore

                try:
                    new_store = GraphStore(self._db_path, read_only=True)
                except Exception as e:
                    self._log.warning("Failed to reopen database: %s", e)
                    # Keep the old store and try again on next access
                    return self._store

            if self._store is not None:
                try:
                    self._store.close()
                except Exception:
                    pass

            self._store = new_store
            self._inode = current_inode
            return self._store

```

agent/src/opentrace_agent/cli/main.py
```diff
@@ -184,28 +185,42 @@ def index(
...
+        store.close()
+    except BaseException:
+        # Clean up the staging file on failure.
+        Path(staging_db).unlink(missing_ok=True)
+        raise
 
-    store.close()
+    # Atomically replace the live database.  Readers holding the old
+    # file descriptor continue to see the previous data until they
+    # detect the inode change and reopen.
```
ghost: ⚠️ This cleanup will fail on Windows if the `graph_store` is still open (even in the error path). It's also better to use a context manager to ensure the store is closed reliably before unlinking the staging file.

Additionally, the message says it's opening `resolved_db`, but it's actually opening `staging_db`.

```suggestion
    click.echo(f"Opening staging database at {staging_db} ...")
    try:
        with GraphStore(staging_db) as graph_store:
            store = GraphStoreAdapter(graph_store, batch_size=batch_size)

            click.echo(f"Indexing {root} ...")
            t0 = time.monotonic()

            inp = PipelineInput(path=str(root), repo_id=repo_id)

            last_result = None
            for event in run_pipeline(inp, store=store):
                _print_event(event, verbose)
                if getattr(event, "result", None) is not None:
                    last_result = event.result

            elapsed = time.monotonic() - t0

            # Save index metadata.
            metadata = _collect_metadata(root, repo_id, elapsed, last_result)
            graph_store.save_metadata(metadata)

            store.flush()
    except BaseException:
        # Clean up the staging file on failure.
        Path(staging_db).unlink(missing_ok=True)
        raise

```

agent/src/opentrace_agent/cli/main.py
```diff
@@ -363,6 +378,67 @@ def stats(db_path: str | None, output_format: str) -> None:
...
+
+    def close(self) -> None:
+        if self._store is not None:
+            self._store.close()
+            self._store = None
+
+
 @app.command("mcp")
 @click.option(
     "--db",
```
ghost, line 475: ⚠️ Added a lock to `close()` as well, for consistency with the `get()` lock.

```suggestion
    def close(self) -> None:
        with self._lock:
            if self._store is not None:
                self._store.close()
                self._store = None

```

agent/src/opentrace_agent/cli/main.py
Umaaz: Acknowledged — OpenTrace currently targets Linux/macOS only (the lock issue itself originates from a unix-style fcntl error message and our docs/CI don't cover Windows). Filing this as a future-Windows concern; for now the fix solves the reported Linux/macOS use case.

agent/src/opentrace_agent/cli/main.py
Umaaz: Fixed in 8a8cf0b7 — added a `threading.Lock` around the inode check + reopen with double-checked locking. I kept the close-then-open ordering instead of the suggested open-then-close: opening the new store before closing the old breaks LadybugDB because the old store still owns the `.wal` file, causing "Database ID does not match" errors on the new open. The test `test_reopens_on_inode_change` caught this.

agent/src/opentrace_agent/cli/main.py
Umaaz: Fixed in 8a8cf0b7 — log message now says `Opening staging database at {staging_db}`, and the GraphStore is opened via `with GraphStore(staging_db) as graph_store:` so close happens reliably before the staging-file unlink in the except path. Inside the context the adapter just calls `store.flush()`; the real close is the context-manager exit.

agent/src/opentrace_agent/cli/main.py
Umaaz: Fixed in 8a8cf0b7 — `close()` now also acquires the lock, consistent with `get()`.

PR #326, opened 2026-04-13
ci: split preview publish into build + workflow_run publish

Umaaz:
## Fix PyPI OIDC publishing for fork PRs
✨ **Improvement** · 🔒 **Security**

Splits the preview release process into two workflows to enable PyPI Trusted Publishing (OIDC) for fork PRs. 

This follows the standard security pattern of building untrusted code in a restricted context and publishing the resulting artifact from a trusted context.

### Complexity
🟢 Low · `2 files changed, 104 insertions(+), 29 deletions(-)`

Narrow scope affecting only two CI workflow files. Follows canonical GitHub Actions patterns for secure artifact handling across trust boundaries.

### Review focus

- **Artifact validation** — verify that the regex guards on the version string and PR number correctly sanitize inputs before they are used in the trusted context.
- **Workflow triggers** — ensure the `workflow_run` condition correctly filters for successful completions of the build workflow only.

### DIFF
--- a/.github/workflows/publish-preview-build.yml
+++ b/.github/workflows/publish-preview-build.yml
@@ -1,0 +1,58 @@
+name: Publish Preview Release Build
+
+# Builds the preview package in the (untrusted) PR context and uploads the
+# resulting dist/ as an artifact. A separate workflow (publish-preview.yml)
+# runs on workflow_run in the base-repo context and handles the PyPI publish.
+#
+# This split is required because PyPI's Trusted Publishing rejects OIDC tokens
+# issued from `pull_request_target` workflows, but the standard `pull_request`
+# trigger from a fork has no OIDC token to issue.
+
+on:
+  pull_request:
+    types: [labeled, synchronize]
+    paths:
+      - "agent/**"
+
+jobs:
+  build:
+    name: Build preview package
+    if: contains(github.event.pull_request.labels.*.name, 'preview-publish')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
+        with:
+          python-version: "3.12"
+
+      - name: Set preview version
+        working-directory: agent
+        run: |
+          BASE_VERSION=$(python3 -c "import tomllib; print(tomllib.load(open('pyproject.toml','rb'))['project']['version'])")
+          PREVIEW_VERSION="${BASE_VERSION}.dev${{ github.event.pull_request.number }}${GITHUB_RUN_NUMBER}"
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-preview-build.yml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  pull_request:
+    types: [labeled, synchronize]
+    paths:
+      - "agent/**"
+
+jobs:
+  build:
+    name: Build preview package
```
ghost, line 19: ⚠️ The `preview-publish` label check here can be bypassed by a contributor by modifying this workflow file in their PR. Since the companion `publish-preview.yml` workflow (which runs in the trusted `main` context) doesn't re-verify the label, this allows any PR to publish to PyPI. 

Consider adding a verification step in `publish-preview.yml` that uses the GitHub API to check the labels of the PR associated with `github.event.workflow_run.head_sha`.

.github/workflows/publish-preview.yml
```diff
@@ -1,54 +1,70 @@
...
-      - name: Build package
-        working-directory: agent
+      - name: Read metadata
+        id: meta
         run: |
-          pip install build
-          python -m build
+          VERSION=$(cat artifact/agent/preview-version.txt)
+          PR_NUMBER=$(cat artifact/agent/pr-number.txt)
+          # Guard against unexpected content in the artifact files.
```
ghost: ⚠️ Relying on the PR number from the artifact is risky as it could be used to post comments to a different PR. It's better to get the PR number directly from the `workflow_run` event or by querying the API using `github.event.workflow_run.head_sha`.

.github/workflows/publish-preview-build.yml
Umaaz: Fixed in b29882de — added a 'Resolve PR and verify label' step in publish-preview.yml that uses listPullRequestsAssociatedWithCommit on workflow_run.head_sha and fails the job unless the matching open PR has the 'preview-publish' label. The build-workflow check is now backed by a trusted-side enforcement that a fork PR can't bypass.

.github/workflows/publish-preview.yml
Umaaz: Fixed in b29882de — the comment step now uses the PR number from the API lookup (steps.resolve_pr.outputs.pr_number) instead of pr-number.txt. Same trusted source as the label check. Also added a check that the version string ends with the trusted .dev{PR_NUM}{BUILD_RUN_NUM} suffix to prevent version squatting via the same artifact-trust hole.

.github/workflows/publish-preview-build.yml
ghost: Nice fix in b29882de! Using the GitHub API as the source of truth for the label gate closes that bypass hole completely. Resolving this one.

.github/workflows/publish-preview.yml
ghost: Excellent addition on the version suffix check in b29882de — that's a solid way to ensure the artifact hasn't been tampered with to squat on other versions. The PR number lookup from the API is also much safer than the artifact. I'll mark this as resolved.

PR #333, opened 2026-04-17
feat(plugin): add debug mode via OPENTRACE_DEBUG

Umaaz:
## Add debug mode with verbose logging for plugin hooks
🆕 **New Feature** · ✨ **Improvement** · 📝 **Docs**

Introduces a verbose debug mode for Claude Code plugin hooks to improve observability. Hooks now log detailed execution traces (commands, durations, exit codes, and output sizes) to stderr and a local `.opentrace/hook-debug.log` file.

- Enable by setting `OPENTRACE_DEBUG=1` before launching Claude.
- Includes a shared `_debug.py` utility for Python hooks and an equivalent Bash implementation for `session-start.sh`.
- Logs are automatically ignored by git via the root `*.log` pattern.
- The session-start `systemMessage` now surfaces the log path when active for easy discovery.

### Complexity
🟢 Low · `5 files changed, 289 insertions(+), 56 deletions(-)`

The changes are localized to the plugin scripts and do not affect the core agent logic. The feature is opt-in and non-breaking for existing hook consumers.

### Tests
🧪 Existing hook integration tests continue to pass; the new debug output is directed to stderr and does not interfere with JSON parsing on stdout.

### Review focus

- **I/O safety** — verify that `OSError` is correctly swallowed in the Python logger and `2>/dev/null` is used in Bash to prevent hook crashes during logging failures.
- **Path discovery** — confirm the upward walk logic in both Python and Bash correctly identifies the `.opentrace` directory and respects the 10-level depth limit.

### DIFF
--- a/claude-code-plugin/scripts/_debug.py
+++ b/claude-code-plugin/scripts/_debug.py
@@ -1,0 +1,116 @@
+# Copyright 2026 OpenTrace Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Shared debug logging helpers for OpenTrace Claude Code plugin hooks.
+
+Enable by setting ``OPENTRACE_DEBUG=1`` in the shell that launches Claude
+Code. When on, hook scripts emit a timestamped trace to stderr *and*, when
+a repo-scoped ``.opentrace/`` directory can be located, append to
+``.opentrace/hook-debug.log`` (alongside the index database).
+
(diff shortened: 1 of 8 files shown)


claude-code-plugin/scripts/_debug.py
```diff
@@ -0,0 +1,116 @@
...
+        """Resolve the log file path once we know the hook's cwd."""
+        if not _DEBUG or self._resolved:
+            return
+        self._log_path = find_log_path(cwd)
+        self._resolved = True
+
+    @property
+    def log_path(self) -> Optional[str]:
+        return self._log_path
+
```
ghost, line 101: ⚠️ The shell script uses UTC (`date -u`), but this Python implementation uses local time by default. For consistency in the same log file, it's better to use UTC here as well.

```suggestion
        line = f"{time.strftime('%Y-%m-%dT%H:%M:%S', time.gmtime())} [{self.tag}] {msg}"
```

Umaaz: Fixed in 6d2c8899 — switched to `time.gmtime()` so Python hooks emit UTC timestamps matching the shell script's `date -u`.

ghost: Looks good! Using `time.gmtime()` in 6d2c8899 ensures consistency with the shell script's UTC timestamps. I'll mark this as resolved.

PR #340, opened 2026-04-21
feat(ui): enforce 500 MB repository archive limit with clear errors

jatinot:
## Enforce 500 MB limit for repository archive downloads
🆕 **New Feature** · ✨ **Improvement**

This PR enforces a 500 MB limit on repository archives to prevent browser-side stalling and memory crashes. 

It adds an upfront notice to the repository modal and implements a streaming size check that aborts the download immediately if the limit is exceeded.

### Complexity
🟢 Low · `3 files changed, 42 insertions(+)`

The scope is limited to the repository fetch utility and a single UI component. It introduces a safeguard using existing progress-reporting patterns without altering core indexing logic.

### Tests
🧪 The new size-limiting logic in the fetch helper is currently untested in the automated suite.

### Review focus

- **Rounding boundary** — verify that `MAX_ARCHIVE_SIZE` correctly triggers exactly when the UI would round the display value to 500.0 MB.
- **Stream cancellation** — ensure `reader.cancel()` effectively halts the network request to prevent further data usage once the limit is hit.

### DIFF
--- a/ui/src/runner/browser/loader/shared.ts
+++ b/ui/src/runner/browser/loader/shared.ts
@@ -26,6 +26,7 @@ import {
   BINARY_EXTENSIONS,
   EXCLUDED_DIRS,
   IMAGE_EXTENSIONS,
+  MAX_ARCHIVE_SIZE,
   MAX_FILE_SIZE,
 } from './constants';
 import type { RepoFile } from '../types';
@@ -221,6 +222,13 @@ export async function fetchResolvedArchive(
 
   const total = contentLength || 0;
 
+  // Reject archives that exceed the size limit
+  if (total >= MAX_ARCHIVE_SIZE) {
+    throw new Error(
+      `Repository exceeds 500 MB limit (${(total / 1024 / 1024).toFixed(0)} MB). Try a smaller repository or use the CLI agent for large codebases.`,
+    );
+  }
+
   // If no ReadableStream body (unlikely in modern browsers), fall back
   if (!zipRes.body) {
     const buf = new Uint8Array(await zipRes.arrayBuffer());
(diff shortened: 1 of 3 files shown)


ui/src/runner/browser/loader/shared.ts
```diff
@@ -221,6 +222,13 @@ export async function fetchResolvedArchive(
...
+  // Reject archives that exceed the size limit
+  if (total >= MAX_ARCHIVE_SIZE) {
+    throw new Error(
+      `Repository exceeds 500 MB limit (${(total / 1024 / 1024).toFixed(0)} MB). Try a smaller repository or use the CLI agent for large codebases.`,
+    );
+  }
+
   // If no ReadableStream body (unlikely in modern browsers), fall back
   if (!zipRes.body) {
     const buf = new Uint8Array(await zipRes.arrayBuffer());
```
ghost, line 234: ⚠️ The size limit check is missing in this fallback path. While rare, if a browser doesn't support streaming or the server doesn't provide a `Content-Length`, a large repository could still be downloaded here and potentially cause an OOM. Consider adding a check on `buf.length` here as well.

ui/src/runner/browser/loader/shared.ts
```diff
@@ -238,7 +246,14 @@ export async function fetchResolvedArchive(
     if (done) break;
     chunks.push(value);
     received += value.length;
+    // Update UI first so the user sees the 500 MB value before the error.
     onProgress?.({ phase: 'download', current: received, total });
+    if (received >= MAX_ARCHIVE_SIZE) {
+      reader.cancel();
+      throw new Error(
+        'Repository exceeds 500 MB limit. Try a smaller repository or use the CLI agent for large codebases.',
+      );
```
ghost, line 260: ⚠️ For consistency with the error message at the start of the function, it's helpful to include the actual size reached in the error message when the limit is hit during streaming.

```suggestion
    if (received >= MAX_ARCHIVE_SIZE) {
      reader.cancel();
      throw new Error(
        `Repository exceeds 500 MB limit (reached ${(received / 1024 / 1024).toFixed(0)} MB). Try a smaller repository or use the CLI agent for large codebases.`,
      );
    }
```

PR #343, opened 2026-04-22
fix(chat): avoid truncating streamed response in finally block

jatinot:
## Fix race condition truncating streamed chat responses
🐛 **Bug Fix**

Fixes an issue where streamed chat responses were randomly truncated mid-sentence. 

Ensures that final usage metadata and message persistence are applied to the absolute latest state snapshot, preventing React 18 batching from overwriting late-arriving chunks.

### Complexity
🟢 Low · `1 file changed, 33 insertions(+), 22 deletions(-)`

The change is localized to a single function in one file and fixes a timing issue using standard React functional update patterns.

### Review focus

- **Functional state update** — verify the transition from raw state writes to a functional setter correctly preserves preceding streamed chunks.
- **Persistence timing** — check that calling `persistMessages` inside the setter is safe despite potential multiple executions by React (the underlying hook uses ref-guards for safety).


ui/src/appComponents/ChatPanel.tsx
```diff
@@ -707,31 +707,42 @@ export default function ChatPanel({
...
-        );
-      }
+        if (persistAfterFlush) {
+          persistMessages(
+            next,
+            providerId,
+            modelId,
+            Array.from(chatFoundNodesRef.current),
+          );
+        }
```
ghost: ⚠️ Side effects like `persistMessages` should not be called inside a state updater function. React updaters are expected to be pure and can be called multiple times. Since `persistMessages` is async and performs its own state updates (e.g., `setConversationId`), calling it here could lead to double-saves or React warnings about updating state during render.

Consider setting a flag in the `finally` block and using a `useEffect` to trigger the persistence once the state has updated.

jatinot: Good catch — fixed in 0214008.

`persistMessages` is now called outside the updater. The usage-merge `setMessages` is wrapped in `flushSync`, which forces React to commit the backlog of per-chunk setters plus our merge synchronously and run effects; after it returns, `messagesRef.current` reflects the final rendered state, so persistence reads from a stable snapshot without running inside a (potentially re-invoked) updater.

dapywell (approved):
Summary
This PR fixes a critical race condition in chat message streaming that was silently truncating responses. The issue arose from React 18's automatic batching: the finally block was reading stale messagesRef.current and performing a non-functional state update that queued after pending per-chunk functional updates, effectively overwriting accumulated stream chunks.

The fix is well-architected and properly handles the complexity of React 18's concurrent features:

Converts to functional updates — setMessages((prev) => ...) ensures the update reads the most current state
Uses flushSync — Forces React to commit all pending updates synchronously before proceeding to persistence
Avoids side effects in updaters — Keeps persistMessages (which calls setConversationId) outside the state updater function
Excellent documentation — The inline comment is comprehensive and explains the React 18 batching subtlety
Issues Found
[Design & Architecture] Severity: Low — ui/src/appComponents/ChatPanel.tsx:741-747

The logic change from nested if to early returns is correct, but there's a subtle behavioral difference. Previously, when hasUsage was false OR when the last message wasn't an assistant message, the code still assigned updatedMessages = messagesRef.current. Now, when those conditions fail, the code returns prev unchanged. This is functionally equivalent since we're inside a flushSync and messagesRef.current would be up-to-date with prev, but it's worth noting for edge cases.


Context: Given that this runs once per stream completion (not per chunk), and the message array is typically small (<100 messages), this is acceptable. However, if users report UI freezes when chat panels have hundreds of messages, this would be the culprit.

Potential optimization: Consider using a simpler pattern if React 19's automatic flushSync behavior arrives, or investigate whether the ref synchronization could be handled differently (though the current solution is solid).

Positive Observations
Excellent problem diagnosis — The comment demonstrates deep understanding of React 18's automatic batching and concurrent rendering implications. This is the kind of documentation that saves future developers hours of debugging.

Proper use of flushSync — This is a legitimate use case for flushSync: ensuring refs are synchronized before running imperative code. The author correctly identified that functional updates alone wouldn't solve the persistence issue.

StrictMode-aware — The comment explicitly mentions StrictMode and concurrent rendering edge cases, showing awareness of React's double-invocation in development.

Defensive programming — The early returns (if (!hasUsage) return prev) avoid unnecessary work and make the code more robust.

Preserves existing behavior — The refactoring maintains the same external behavior while fixing the truncation bug.

PR #344, opened 2026-04-22
feat(ui): add regenerate button to assistant messages

jatinot:
## Add regenerate button and fix message role type narrowing
🆕 **New Feature** · 🐛 **Bug Fix**

This PR adds a "Regenerate" button to the final assistant message in a thread. It also fixes a TypeScript build error caused by lost type narrowing when indexing into the message array.

Clicking the button aborts any active stream, removes the current turn, and re-submits the previous user prompt (including attachments) to get a fresh response.

### Complexity
🟢 Low · `2 files changed, 75 insertions(+), 7 deletions(-)`

Introduces a straightforward UI feature by leveraging existing message truncation and resending patterns. The fix is a simple type guard to resolve a compiler limitation.

### Review focus

- **Message Truncation** — verify that `regenerateMessage` correctly slices the array to remove both the old user and assistant messages before resending.
- **Stream Abortion** — confirm that `abortRef` is triggered correctly if a user regenerates while a response is still in progress.

### DIFF
--- a/ui/src/appComponents/ChatPanel.tsx
+++ b/ui/src/appComponents/ChatPanel.tsx
@@ -794,6 +794,7 @@ export default function ChatPanel({
     }
     if (userIdx < 0) return;
     const userMsg = messages[userIdx];
+    if (userMsg.role !== 'user') return;
     if (streaming) {
       abortRef.current?.abort();
       streamingRef.current = false;
--- a/ui/src/appComponents/ChatPanel.tsx
+++ b/ui/src/appComponents/ChatPanel.tsx
@@ -784,6 +784,26 @@ export default function ChatPanel({
     setEditText('');
     await sendMessage(trimmed, original.images, original.attachments);
   };
+  const regenerateMessage = async (assistantIndex: number) => {
+    let userIdx = -1;
+    for (let k = assistantIndex - 1; k >= 0; k--) {
+      if (messages[k].role === 'user') {
+        userIdx = k;
+        break;
+      }
+    }
+    if (userIdx < 0) return;
+    const userMsg = messages[userIdx];
+    if (streaming) {
+      abortRef.current?.abort();
+      streamingRef.current = false;
+      setStreaming(false);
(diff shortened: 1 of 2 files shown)


ui/src/appComponents/ChatPanel.tsx
```diff
@@ -784,6 +784,27 @@ export default function ChatPanel({
...
     await sendMessage(trimmed, original.images, original.attachments);
   };
+  const regenerateMessage = async (assistantIndex: number) => {
+    let userIdx = -1;
+    for (let k = assistantIndex - 1; k >= 0; k--) {
+      if (messages[k].role === 'user') {
+        userIdx = k;
+        break;
+      }
+    }
```
bedonath, line 794: Maybe worth extracting a little helper function for this as the same code is already used further down (line 1316)

ui/src/appComponents/ChatPanel.tsx
```diff
@@ -784,6 +784,27 @@ export default function ChatPanel({
...
+    if (userIdx < 0) return;
+    const userMsg = messages[userIdx];
+    if (userMsg.role !== 'user') return;
+    if (streaming) {
```
bedonath, line 798: This block is unreachable as the code stands today:

```
if (streaming) {
  abortRef.current?.abort();
  streamingRef.current = false;
  setStreaming(false);
}
```

The render path gates the regenerate button on both `i === messages.length - 1` and `!(streaming && i === messages.length - 1)`. Combined, that means the button is only
visible when `streaming === false` for the last message — so by the time regenerateMessage runs, streaming is guaranteed to be false and the branch never executes.

That contradicts the PR description, which states:

▎ Clicking the button aborts any active stream, removes the current turn, and re-submits the previous user prompt.

…and the review-focus note:

▎ Stream Abortion — confirm that abortRef is triggered correctly if a user regenerates while a response is still in progress.

bedonath (approved):

PR #347, opened 2026-04-22
feat(ui): add reindex-repo job to rebuild a single repo's graph

AnysSido:
## Add single-repo reindexing and fix import PK collisions
🆕 **New Feature** · ✨ **Improvement** · 🐛 **Bug Fix** · 🧪 **Tests**

Adds a `reindex-repo` job type to rebuild a single repository's graph without wiping the entire store. 

Fixes a latent bug where importing a database led to primary-key collisions on shared dependency nodes during subsequent indexing runs.

### Complexity
🟡 Moderate · `5 files changed, 416 insertions(+), 1 deletion(-)`

This PR modifies the core state management logic in `LadybugGraphStore`, requiring precise synchronization between the WASM-based database and multiple in-memory caches (BM25, LRU caches, source snippets). It introduces a targeted deletion strategy that must respect database referential integrity while navigating the project's node ID conventions.

### Tests
🧪 Includes thorough unit tests verifying Cypher query ordering, repo-scoped filtering, and in-memory state pruning.

### Review focus

- **Deletion Ordering** — verify that `RELATES` edges are dropped before node tables to satisfy database referential integrity.
- **ID Predicates** — confirm the `MATCH` clauses correctly isolate repo-scoped nodes from global `Dependency` nodes and other repos.
- **State Pruning** — ensure all JS-side indexes and buffers are correctly filtered to prevent stale data from leaking into the new index.

---

<details>

### Approach
The re-index capability is implemented via a new `deleteRepo` method on the `GraphStore` interface.

**Repo Isolation**
The logic relies on the convention that repository-scoped nodes use IDs prefixed with `${repoId}/` (e.g., `owner/repo/path/to/file`). The `deleteRepo` method uses Cypher predicates to target these IDs and the repository node itself, while specifically excluding `Dependency` nodes which are global (`pkg:registry:name`) and shared across repositories.

**Synchronization**
Unlike `clearGraph`, `deleteRepo` does not bump the `generation` counter. This allows queries for other repositories to remain valid while one repo is being cleared. The store's internal serialization queue ensures the deletion completes before new writes for that repo begin.

**Fix for Import Collisions**
The `importDatabase` path previously skipped updating `flushedPackageIds` because it performed a bulk `COPY FROM` from Parquet. This led to uniqueness failures if a subsequent index job tried to re-emit the same package nodes. Both `deleteRepo` and `importDatabase` now invoke `rebuildPackageDedupIndex` to reconcile the in-memory guard with ground-truth DB rows.

</details>

### DIFF
--- a/ui/src/store/ladybugStore.ts
+++ b/ui/src/store/ladybugStore.ts
@@ -1503,6 +1503,136 @@ export class LadybugGraphStore implements GraphStore {
     this.totalRelsBuffered = 0;
   }
 
+  /** Remove all data scoped to a single repo: the Repository node itself,
+   *  every node whose ID starts with `${repoId}/` (Directory, File, Class,
+   *  Function, Variable, PullRequest, SourceText, NodeVector), the
+   *  IndexMetadata row, and every RELATES edge touching them. Only
+   *  Dependency nodes are truly global (`pkg:registry:name`) and survive.
+   *  Derived JS-side indexes are pruned in lockstep.
+   *
+   *  Unlike clearGraph() this does NOT bump the generation counter: other
+   *  repos' in-flight queries remain valid. The store's serialization queue
+   *  ensures our deletes run between neighboring queries. */
+  async deleteRepo(repoId: string): Promise<void> {
+    if (!repoId) return;
+    await this.ensureReady();
+
+    const prefix = `${repoId}/`;
+    const metaId = `_meta:index:${repoId}`;
+    const matches = (id: string): boolean =>
+      id === repoId || id.startsWith(prefix) || id === metaId;
+
+    // Drop any buffered writes that would re-create this repo's rows on the
+    // next flush — they'd collide with the deletions we're about to do.
+    if (this.pendingNodes.length) {
+      const before = this.pendingNodes.length;
+      this.pendingNodes = this.pendingNodes.filter((n) => !matches(n.id));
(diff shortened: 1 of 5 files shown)


ui/src/store/ladybugStore.ts
```diff
@@ -1503,6 +1503,136 @@ export class LadybugGraphStore implements GraphStore {
...
+   *      never touching the set.
+   *  Both call this afterwards to bring the set back in sync. Cheap —
+   *  one id column, bounded by package count, not repo count. */
+  private async rebuildPackageDedupIndex(): Promise<void> {
+    this.flushedPackageIds.clear();
+    const depRows = (await this.query(
+      `MATCH (n:Dependency) RETURN n.id AS id`,
+    )) as { id: string }[];
+    for (const row of depRows) this.flushedPackageIds.add(row.id);
+  }
```
ghost, line 1634: ⚠️ The LadybugDB WASM layer can return boxed `String` objects instead of primitives. If `row.id` is boxed, `flushedPackageIds.has(node.id)` might fail during the next import because it performs a strict equality check. Wrapping it in `unbox` ensures we store primitive strings.

```suggestion
    for (const row of depRows) this.flushedPackageIds.add(unbox(row.id) as string);

```

brandontaslater (approved):

PR #353, opened 2026-04-24
fix(ui): keep graph framed during indexing, not only after embeddings (OT-1712)

bedonath:
## Auto-frame graph during indexing and disable existing example repos
✨ **Improvement** · 🐛 **Bug Fix** · 🧪 **Tests**

Implements dynamic "zoom-to-fit" tracking during the indexing process so nodes stay framed as the layout evolves. Also improves the repository picker by disabling example repo chips that are already indexed in the current project.

### Complexity
🟡 Moderate · `4 files changed, 88 insertions(+), 5 deletions(-)`

This PR touches the core rendering loop and interaction state of the Pixi-based graph. It requires coordinating state between the React component tree (GraphViewer) and the imperative renderer (PixiRenderer) to manage throttled animations and user-driven state overrides.

### Tests
🧪 Includes a new unit test for the repository validation logic in AddRepoModal.

### Review focus

- **Animation race conditions** — check that `cancelAnimation` and the throttle timer in `PixiRenderer` correctly handle rapid updates from the force simulation.
- **Interaction state** — verify that manual pan/zoom correctly overrides and sticks, preventing the "auto-fit" from hijacking the camera back after user movement.

### DIFF
--- a/ui/src/components/pixi/PixiRenderer.ts
+++ b/ui/src/components/pixi/PixiRenderer.ts
@@ -220,8 +220,9 @@ export class PixiRenderer {
   private height = 0;
 
   // Auto-fit state — keep the graph framed during indexing until user takes
-  // control. Flipped to true by wheel/pan/rotate/zoomToNodes, reset by
-  // setData() (new graph session) and resetCamera() (explicit "Reset View").
+  // control. Flipped to true by any gesture past CLICK_THRESHOLD (wheel,
+  // pan, 3D rotate, node drag) and by zoomToNodes; reset by setData() (new
+  // graph session) and resetCamera() (explicit "Reset View").
   private hasUserMovedCamera = false;
   private autoFitThrottleTimer: ReturnType<typeof setTimeout> | null = null;
   private readonly AUTO_FIT_THROTTLE_MS = 180;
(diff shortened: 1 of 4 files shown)


ui/src/components/pixi/PixiRenderer.ts
```diff
@@ -1868,6 +1911,8 @@ export class PixiRenderer {
               this.mode3dAutoRotate = false;
               this.callbacks.on3DAutoRotateChange?.(false);
```
ghost, line 1918: ⚠️ This flag should also be set when a node drag starts. If a user is manually repositioning a node while the simulation is still running, an auto-fit jump could be very jarring.

Since any interaction beyond a simple click should probably count as manual control, you might want to set this at the start of the `if (movedDistance > CLICK_THRESHOLD)` block to cover node dragging, panning, and rotation in one place.

```suggestion
            // User took manual control — stop auto-fitting during indexing.
            this.hasUserMovedCamera = true;
            this.debouncedLabelCull();
```

bedonath: Good catch — applied in a74f35f. Hoisted `hasUserMovedCamera = true` to the start of the `movedDistance > CLICK_THRESHOLD` block so node drag, pan, and 3D rotate all mark manual control in one place. Also updated the field comment to reflect the full trigger list.

ghost: Nicely handled in a74f35fb. Hoisting the flag to the threshold check correctly captures node drags as manual control, ensuring the auto-fit doesn't fight the user while they're repositioning nodes. I'll mark this as resolved.

AnysSido (approved):

PR #354, opened 2026-04-24
feat(ui): resizable graph panels — horizontal, vertical, corner (OT-1716)

Edwinbhart:
## Unified resizable panels with horizontal, vertical, and corner handles
🆕 **New Feature** · ✨ **Improvement** · ♻️ **Refactor** · ⚡ **Performance**

Consolidates bespoke panel resizing into a shared `PanelResizeHandle` component and high-performance hooks. 

Adds vertical and corner resizing to several panels while optimizing drag interactions to maintain responsiveness in heavy virtualized trees.

### Complexity
🟡 Moderate · `16 files changed, 584 insertions(+), 85 deletions(-)`

This PR introduces a new shared architectural pattern for UI interactions across the entire application. It moves away from standard React state updates for real-time resizing in favour of a performance-optimised approach using direct DOM manipulation and `requestAnimationFrame`. 

While the change is contained to the UI layer, it touches almost every major panel component and establishes foundational primitives (`PanelResizeHandle` and the companion hooks) that future panels will build on.

### Tests
🧪 Manual verification is required for drag interactions across different viewport sizes as these changes bypass standard React state during the interaction.

### Review focus

- **Direct DOM manipulation** — ensure `useResizablePanel` correctly commits the final width/height to React state on `mouseup` and clears all references to avoid memory leaks.
- **CSS `:has()` compatibility** — verify that the project's browser support targets allow the `:has()` selector used for cascading hover effects on corners.
- **Z-index and layout** — check that the new 10px insets on resize handles don't interfere with inner panel content or create clickable dead zones.

### DIFF
--- a/ui/src/hooks/useResizablePanel.ts
+++ b/ui/src/hooks/useResizablePanel.ts
@@ -101,9 +101,19 @@ export function useResizablePanel({
 
     const flush = () => {
       rafId = null;
-      if (pending != null && panelRef.current) {
-        panelRef.current.style.width = `${pending}px`;
+      const panel = panelRef.current;
+      if (pending == null || !panel) return;
+      // Clamp pending to whatever the parent currently allows so the
+      // drag stops cleanly at the available limit (and doesn't fight the
+      // ResizeObserver's clamp on every frame).
+      const parent = panel.parentElement;
+      if (parent) {
+        const panelRect = panel.getBoundingClientRect();
+        const parentRect = parent.getBoundingClientRect();
+        const available = parentRect.right - panelRect.left - 4;
+        if (available > 0 && pending > available) pending = available;
       }
+      panel.style.width = `${pending}px`;
     };
 
     const handleMouseMove = (e: MouseEvent) => {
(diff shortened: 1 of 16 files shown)


ui/src/hooks/useResizablePanel.ts
```diff
@@ -49,51 +57,183 @@ export function useResizablePanel({
...
+      isDragging.current = true;
+      startY.current = e.clientY;
+      // Anchor the drag to whatever height the panel is ACTUALLY rendering —
+      // this matters most before first drag when `height` is still null.
+      const panel = (e.currentTarget as HTMLElement).parentElement;
+      panelEl.current = panel;
+      const rendered = panel?.getBoundingClientRect().height ?? 0;
+      startHeight.current = height ?? Math.round(rendered);
+      document.body.style.cursor = 'row-resize';
+      document.body.style.userSelect = 'none';
```
ghost: ⚠️ When dragging a corner handle, both horizontal and vertical hooks are triggered. They both set `document.body.style.cursor`, causing a race where one axis's cursor (e.g. `row-resize`) overrides the other, rather than showing a diagonal cursor.

Consider adding a `cursor` option to the hooks so the caller can provide the correct diagonal cursor (like `nwse-resize`) when a corner handle is used.

ui/src/hooks/useResizablePanel.ts
```diff
@@ -49,51 +57,183 @@ export function useResizablePanel({
...
+
+  const handleMouseDown = useCallback(
+    (e: React.MouseEvent) => {
+      e.preventDefault();
+      isDragging.current = true;
+      startY.current = e.clientY;
+      // Anchor the drag to whatever height the panel is ACTUALLY rendering —
+      // this matters most before first drag when `height` is still null.
+      const panel = (e.currentTarget as HTMLElement).parentElement;
+      panelEl.current = panel;
```
ghost: ⚠️ This assumes that the resize handle is a direct child of the panel being resized. This makes the hook fragile if the DOM structure changes (e.g. if the handle is wrapped in another container).

A more robust approach would be to pass the panel's `RefObject` directly to the hook.

ui/src/hooks/useResizablePanel.ts
```diff
@@ -49,51 +57,183 @@ export function useResizablePanel({
...
   const startWidth = useRef(0);
+  const panelEl = useRef<HTMLElement | null>(null);
 
   const handleMouseDown = useCallback(
     (e: React.MouseEvent) => {
       e.preventDefault();
       isDragging.current = true;
       startX.current = e.clientX;
       startWidth.current = width;
+      panelEl.current = (e.currentTarget as HTMLElement).parentElement;
```
ghost: ⚠️ Same issue here: the `.parentElement` assumption is brittle. Passing a ref to the panel would be more reliable.

ui/src/hooks/useResizablePanel.ts
```diff
@@ -49,51 +57,183 @@ export function useResizablePanel({
...
       document.body.style.cursor = 'col-resize';
```
ghost: ⚠️ This cursor will override the vertical hook's cursor (or vice-versa) during a corner drag. Providing an optional `cursor` prop to the hook would allow the caller to set the appropriate diagonal cursor for corners.

brandontaslater (changes requested):
@Edwinbhart - if you resize the browser window - i notice the resize corner grabber is no longer visible - do you get this artefact? probably something to do with storing the height in browser storage (maybe its cached/ and doesn't dynamically adjust based on window resize)

ui/src/hooks/useResizablePanel.ts
ghost: I see you've added the `cursorOverride` parameter to the `handleMouseDown` callback in 2526c3e. Passing `nwse-resize` from the corner handles in `SidePanel` and `PhysicsPanel` resolves the race condition and ensures the correct diagonal cursor is shown. Looks good! I'll mark this as resolved.

ui/src/hooks/useResizablePanel.ts
ghost: Passing the `panelRef` directly to the hook is much cleaner than relying on the DOM hierarchy. Nice fix in 2526c3e. Closing this out.

ui/src/hooks/useResizablePanel.ts
ghost: The `cursorOverride` approach in 2526c3e handles this nicely. Resolving.

ui/src/hooks/useResizablePanel.ts
ghost: Addressed in 2526c3e by introducing `panelRef` to the vertical hook as well. Resolving this one.

brandontaslater (approved):

PR #374, opened 2026-04-28
feat(agent): add MCP tools for source access, search, usages, impact, and remote indexing

jatinot:
## Add MCP tools, remote indexing, and hot-reload support
🆕 **New Feature** · ✨ **Improvement** · 🐛 **Bug Fix** · 🧪 **Tests**

Adds a comprehensive suite of MCP tools for agents to search, read, and analyze codebases alongside support for remote repository indexing. It stabilizes the indexing pipeline to prevent relationship data loss and implements a hot-reload strategy for concurrent indexing and querying.

### Complexity
🟠 High · `12 files changed, 1414 insertions(+), 72 deletions(-)`

This PR spans the entire stack, modifying the low-level database adapter, the indexing pipeline logic, and the high-level MCP tool definitions. It introduces a new architectural pattern for database hot-reloading and handles remote repository state management, which significantly increases the system's complexity and statefulness.

### Tests
🧪 Includes thorough integration tests for all MCP tools and verifies the hot-reload database proxy using real fixtures.

### Note
⚠️ The indexing pipeline now strictly enforces a node-before-relationship flush order. If this is accidentally reverted in future refactors, relationships will be silently dropped by the underlying Cypher engine.

### Review focus

- **Indexing atomicity** — verify the `_ReloadableStore` proxy correctly handles inode changes without leaking file descriptors.
- **Relationship integrity** — check the `GraphStoreAdapter` flush logic to ensure endpoint nodes are always persisted before their edges.
- **Subprocess safety** — review the `repo_index` and `source_grep` tools for safe handling of shell commands and user-provided paths.

---

<details>

### Database Concurrency & Hot-Reload
LadybugDB (backed by Kuzu) uses exclusive file locking, which previously prevented the `index` CLI from running while the MCP server was active. This PR implements a staging-and-swap strategy:
* The indexer writes to `index.db.staging`.
* Once complete, it uses `os.replace` to atomically move the staging file over the live database.
* The `_ReloadableStore` proxy detects this change via inode monitoring and transparently reopens the database without dropping the MCP connection.

### Pipeline Reliability
We identified a significant data loss issue where approximately 15% of relationships were missing from indexes. This occurred because the `GraphStoreAdapter` would occasionally flush relationship buffers while the corresponding endpoint nodes were still sitting in the node buffer. Since Cypher's `MATCH...CREATE` pattern silently fails if endpoints don't exist, these edges were lost. The adapter now enforces a strict "flush nodes before relationships" invariant.

### Search Tokenization
The `keyword_search` tool addresses a common failure mode in FTS where natural language queries (e.g., "functions that handle auth") fail to match literal symbol names. It tokenizes the query, strips common English stopwords and LLM-injected filler nouns (like "code" or "function"), and performs a merged, ranked search across the remaining high-signal keywords.

</details>

### DIFF
--- a/agent/src/opentrace_agent/cli/mcp_server.py
+++ b/agent/src/opentrace_agent/cli/mcp_server.py
@@ -816,11 +816,33 @@ def create_mcp_server(
             if not file_nodes:
                 return json.dumps({"error": f"File not found in index: {target}"})
 
-            # Prefer an exact path-suffix match over the FTS top result.
-            file_node = file_nodes[0]
-            for fn in file_nodes:
+            # search_nodes matches on name / search_text / docs, not just
+            # the path — so a query that hits an unrelated File's
+            # docstring would otherwise misroute. Require the chosen
+            # node's path to actually correspond to the target.
+            target_bn = os.path.basename(target)
+
+            def _path_matches(fn: dict[str, Any]) -> bool:
+                node_path = (fn.get("properties") or {}).get("path") or ""
+                if not node_path:
+                    return False
+                if node_path == target or node_path.endswith(target) or target.endswith(node_path):
+                    return True
+                if target_bn:
+                    node_bn = os.path.basename(node_path)
+                    if node_bn == target_bn or node_path.endswith("/" + target_bn):
+                        return True
+                return False
+
+            matched = [fn for fn in file_nodes if _path_matches(fn)]
+            if not matched:
+                return json.dumps({"error": f"File not found in index: {target}"})
+
+            # Prefer an exact path or path-suffix match over a basename-only match.
+            file_node = matched[0]
+            for fn in matched:
                 node_path = (fn.get("properties") or {}).get("path", "")
-                if node_path.endswith(target) or target.endswith(node_path):
+                if node_path == target or node_path.endswith(target) or target.endswith(node_path):
                     file_node = fn
                     break
 
--- a/agent/src/opentrace_agent/cli/mcp_server.py
+++ b/agent/src/opentrace_agent/cli/mcp_server.py
@@ -621,15 +621,29 @@ def create_mcp_server(
 
             repos = _repo_paths(store)
             candidates: list[str] = []
+
+            # Resolve file_path under each repo root and reject any result
+            # that escapes the root (absolute file_path discards the root,
+            # and `..` segments traverse out of it).  Resolving both sides
+            # also makes the comparison symlink-safe.
+            def _safe_join(abs_path: str) -> str | None:
+                root = Path(abs_path).resolve()
+                joined = (root / file_path).resolve()
+                if joined.is_relative_to(root):
+                    return str(joined)
+                return None
+
             if repo_hint:
                 for rid, abs_path in repos:
                     if rid == repo_hint:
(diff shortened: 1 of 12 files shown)


agent/src/opentrace_agent/cli/mcp_server.py
```diff
@@ -182,4 +472,431 @@ def get_stats() -> str:
...
+            candidates: list[str] = []
+            if repo_hint:
+                for rid, abs_path in repos:
+                    if rid == repo_hint:
+                        candidates.append(os.path.join(abs_path, file_path))
+                        break
+            for _, abs_path in repos:
+                joined = os.path.join(abs_path, file_path)
+                if joined not in candidates:
+                    candidates.append(joined)
```
ghost: 🚨 This is a path traversal vulnerability. `os.path.join` will discard `abs_path` if `file_path` is an absolute path, and `../../` can be used to escape the directory. 

Consider resolving the path and checking if it starts with the repository's absolute path:
```python
joined = Path(abs_path).joinpath(file_path).resolve()
if joined.is_relative_to(abs_path):
    candidates.append(str(joined))
```

agent/src/opentrace_agent/cli/mcp_server.py
```diff
@@ -182,4 +472,431 @@ def get_stats() -> str:
...
+
+        try:
+            limit = max(1, min(limit, 1000))
+            results: list[str] = []
+            total = 0
+            for repo_id, abs_path in repos:
+                cmd = [rg, "--no-heading", "--line-number", "--max-count", str(limit)]
+                if include:
+                    cmd.extend(["--glob", include])
+                cmd.extend([pattern, abs_path])
```
ghost: ⚠️ To prevent `pattern` or `abs_path` from being interpreted as flags if they start with a hyphen, use the `--` separator.

```suggestion
                cmd.extend(["--", pattern, abs_path])
```

agent/src/opentrace_agent/cli/mcp_server.py
jatinot: Fixed in 78edc6b — resolved file_path under each repo root and added an is_relative_to(root) check; absolute paths, `..` traversal, and symlink escapes are now rejected before being added as candidates. Both root and joined path are .resolve()d so the check is symlink-safe on both sides.

agent/src/opentrace_agent/cli/mcp_server.py
jatinot: Fixed in 78edc6b — applied the suggestion verbatim, `cmd.extend(["--", pattern, abs_path])`.

agent/src/opentrace_agent/cli/mcp_server.py
ghost: The `_safe_join` helper in 78edc6b correctly uses `Path.resolve()` and `is_relative_to` to prevent traversal. Resolving this one.

agent/src/opentrace_agent/cli/mcp_server.py
ghost: Nice, 78edc6b applies the `--` separator as suggested. Closing this out.

PR #375, opened 2026-04-28
feat(plugin): consolidate plugins under plugins/ and add Codex CLI plugin

jatinot:
## Standardize plugin structure and add Codex CLI integration
🆕 **New Feature** · ♻️ **Refactor** · 📝 **Docs** · 🧪 **Tests**

Consolidates agent plugins under a new `plugins/` directory and introduces a full integration for the Codex CLI. 

This includes a new Codex plugin bundle (MCP + Skills) and a set of runtime hooks that augment shell commands with graph context.

Standardizes the project structure by moving `claude-code-plugin/` to `plugins/claude-code/` and updating all internal references.

### Complexity
🟡 Moderate · `52 files changed, 1476 insertions(+), 13 deletions(-)`

This PR standardizes the project's plugin architecture by moving the Claude Code integration and adding a comprehensive new integration for the Codex CLI. While the Codex integration follows patterns established by the Claude plugin, it introduces a significant volume of new code, including a standalone installer and a runtime hook system that intercepts shell commands. Reviewers need to verify path updates across benchmarks, documentation, and test fixtures.

### Tests
🧪 Existing benchmark and CLI test fixtures have been updated to reflect the new directory structure.

### Review focus

- **Hook performance** — verify that the 7-second timeout and fail-closed logic in `plugins/codex/.codex/hooks/common.py` prevent the `PreToolUse` hook from blocking the Codex CLI during failures.
- **Shell parsing** — check the `extract_search_pattern` and `extract_read_path` logic in `common.py` to ensure it correctly handles standard `rg`, `grep`, and `cat` invocations without false positives.
- **Refactored paths** — ensure all benchmark and documentation references to the old `claude-code-plugin/` path have been correctly updated to `plugins/claude-code/`.


plugins/codex/.agents/plugins/marketplace.json
```diff
@@ -0,0 +1,20 @@
...
+  "name": "opentrace-oss",
+  "interface": {
+    "displayName": "OpenTrace"
+  },
+  "plugins": [
+    {
+      "name": "opentrace",
+      "source": {
+        "source": "local",
+        "path": "./plugins/opentrace"
```
ghost, line 11: 🚨 Should this be `../../plugins/opentrace`? Most marketplace manifests resolve paths relative to the manifest file itself. If so, `./plugins/opentrace` will look in `.agents/plugins/plugins/opentrace`, which doesn't exist.

plugins/codex/plugins/opentrace/.mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mcpServers": {
+    "opentrace": {
+      "command": "opentraceai",
+      "args": ["mcp"]
```
ghost: ⚠️ This will fail if the user doesn't have `opentraceai` installed globally in their PATH. For consistency with the Claude Code plugin and ease of use, it should use `uvx`.

```suggestion
      "command": "uvx",
      "args": ["opentraceai", "mcp"]
```

plugins/codex/.codex/hooks/common.py
```diff
@@ -0,0 +1,250 @@
...
+import shlex
+import shutil
+import subprocess
+import sys
+import time
+from pathlib import Path
+from typing import Optional
+
+
+CACHE_DIR = Path(os.environ.get("TMPDIR", "/tmp")) / "opentrace-codex-hooks"
```
ghost: ⚠️ Using a hardcoded path in `/tmp` can cause permission errors or collisions on multi-user systems. It's safer to include the user ID in the path.

```suggestion
CACHE_DIR = Path(os.environ.get("TMPDIR", "/tmp")) / f"opentrace-codex-hooks-{os.getuid()}"
```

plugins/codex/.codex/hooks/pre_tool_use.py
```diff
@@ -0,0 +1,50 @@
...
+
+    message = None
+    pattern = extract_search_pattern(command)
+    if pattern:
+        message = build_search_message(pattern, workspace_root)
+    else:
+        path = extract_read_path(command)
+        if path:
+            message = build_read_message(path, workspace_root)
+
```
ghost, line 51: ⚠️ Shell commands often use paths relative to the current working directory, not the workspace root. We should resolve the path against the event's `cwd` before calling `build_read_message`.

```suggestion
    else:
        path = extract_read_path(command)
        if path:
            cwd = event.get("cwd") or "."
            abs_path = os.path.abspath(os.path.join(cwd, path))
            message = build_read_message(abs_path, workspace_root)
```

plugins/codex/plugins/opentrace/README.md
```diff
@@ -0,0 +1,162 @@
...
+  auto-discovers from any subdirectory.
+
+## Install
+
+See the parent [`../../README.md`](../../README.md) for the full install
+walkthrough. Quick version:
+
+```bash
+# 1. Register the marketplace (from an OpenTrace checkout)
+codex plugin marketplace add ./codex-plugin
```
ghost: ⚠️ This path is incorrect given the repository structure. It should be `./plugins/codex` if adding from the root.

```suggestion
codex plugin marketplace add ./plugins/codex
```

plugins/codex/plugins/opentrace/.mcp.json
jatinot: Fixed in ec91750 — reverted to `uvx opentraceai` per the suggestion; matches the Claude Code plugin's wiring and works for users without a global install.

plugins/codex/.codex/hooks/common.py
jatinot: Fixed in ec91750 — appended `os.getuid()` to the cache dir, with a Windows fallback (`getattr(os, 'getuid', lambda: 'shared')`) since Codex CLI is unix-first today but no reason to crash on Windows.

plugins/codex/.codex/hooks/pre_tool_use.py
jatinot: Fixed in ec91750 — the hook now resolves shell read paths against `event['cwd']` (with absolute-path passthrough) before calling `build_read_message`. Good catch — `cat src/foo.py` from a subdir was pointing at the wrong file.

plugins/codex/plugins/opentrace/README.md
jatinot: Fixed in ec91750 — stale reference from before the `plugins/` consolidation; updated to `./plugins/codex`.

plugins/codex/.agents/plugins/marketplace.json
jatinot: Not changing — our path is correct per the openai/codex marketplace spec ([source](https://github.com/openai/codex/blob/main/codex-rs/skills/src/assets/samples/plugin-creator/references/plugin-json-spec.md)):

> `path` (`string`): Relative plugin path based on the marketplace root.
>   - Repo plugin: `./plugins/<plugin-name>`
>   - Local plugin in `~/.agents/plugins/marketplace.json`: `./plugins/<plugin-name>`
>   - Example: with `~/.agents/plugins/marketplace.json`, `./plugins/<plugin-name>` resolves to `~/plugins/<plugin-name>`.

Paths are resolved against the **marketplace root** (the parent of `.agents/`), not the manifest file. So with our manifest at `plugins/codex/.agents/plugins/marketplace.json`, the marketplace root is `plugins/codex/`, and `./plugins/opentrace` resolves to `plugins/codex/plugins/opentrace/` — which is exactly where the plugin lives.

Verified end-to-end with `codex plugin marketplace add ./plugins/codex` followed by `/plugins` — the plugin shows up and registers a `[plugins."opentrace@opentrace-oss"]` stanza in `~/.codex/config.toml`.

ghost: Thanks for the clarification on the marketplace spec. If the path is resolved against the marketplace root rather than the manifest itself, then `./plugins/opentrace` is correct. I'll mark this as resolved.

plugins/codex/.codex/hooks/pre_tool_use.py
ghost: Resolving this one—the explicit `cwd` resolution in ec91750a correctly handles relative paths from subdirectories.

plugins/codex/plugins/opentrace/README.md
ghost: Nice catch on the stale path; the update to `./plugins/codex` in ec91750a looks correct now. Resolving.

plugins/codex/plugins/opentrace/.mcp.json
ghost: Looks good, switching to `uvx` in ec91750a ensures the MCP server works out of the box without requiring a global install. Resolving this.

plugins/codex/.codex/hooks/common.py
ghost: The addition of `os.getuid()` in ec91750a is exactly what I was looking for to prevent cache collisions. Good call on the Windows fallback too. I'll close this out.

PR #376, opened 2026-04-28
refactor(ui): lift graph state into context providers

brandontaslater:
## Refactor graph state into centralized context providers
♻️ **Refactor** · ✨ **Improvement**

Moves graph data, selection, and filtering logic out of `GraphViewer` into centralized context providers. This decouples the canvas from side panels and allows downstream components to consume state without prop-drilling.

### Complexity
🟡 Moderate · `8 files changed, 1106 insertions(+), 677 deletions(-)`

This PR performs a large-scale refactor of the application's core state, moving the primary graph data and interaction logic out of a monolithic component and into a provider layer. While it uses standard React patterns (Context + useMemo), the scope is broad and touches nearly every major UI component, meaning a bug here could break the entire navigation and filtering loop.

### Tests
🧪 Existing tests for `useGraphData` and `useHighlights` still apply, but the component-level interaction logic is currently untested.

### Review focus

- **Selection Helpers** — verify that the `selectNode` and `selectLink` helpers in the provider correctly clear conflicting states (like community focus).
- **History Duplication** — check that node selection history is recorded exactly once now that logic exists in both the provider and the viewer's click handler.
- **Memoization Stability** — ensure `filteredGraphData` and `highlights` only re-calculate when necessary to avoid performance lag on large graphs.
- **Layout Calculations** — confirm that moving `SidePanel` to a sibling of `GraphViewer` in `App.tsx` preserves correct canvas resizing logic.

### DIFF
--- a/ui/src/appComponents/GraphViewer.tsx
+++ b/ui/src/appComponents/GraphViewer.tsx
@@ -508,7 +508,6 @@ const GraphViewer = memo(
         setHiddenNodeTypes,
         hiddenLinkTypes,
         hiddenSubTypes,
-        setHiddenSubTypes,
         hiddenCommunities,
         setHiddenCommunities,
         colorMode,
@@ -528,8 +527,6 @@ const GraphViewer = memo(
       const [showResetConfirm, setShowResetConfirm] = useState(false);
       const [showExportModal, setShowExportModal] = useState(false);
       const [exporting, setExporting] = useState(false);
-      // Track whether we've applied the default Dependency hiding
-      const defaultsApplied = useRef(false);
 
       // Append chat-highlighted nodes to session history
       useEffect(() => {
@@ -728,21 +725,6 @@ const GraphViewer = memo(
         [repulsion],
       );
 
-      // On first data load, hide all Dependency sub-types by default
-      useEffect(() => {
-        if (defaultsApplied.current) return;
-        const depSubs = availableSubTypes.get('Dependency');
-        if (depSubs && depSubs.length > 0) {
-          defaultsApplied.current = true;
-          // eslint-disable-next-line react-hooks/set-state-in-effect -- intentional one-time default init
-          setHiddenSubTypes((prev) => {
-            const next = new Set(prev);
-            depSubs.forEach((s) => next.add(`Dependency:${s.subType}`));
-            return next;
-          });
-        }
-      }, [availableSubTypes, setHiddenSubTypes]);
-
       const onNodeClick = useCallback(
         (node: GraphNode) => {
           setSelectedNode(node);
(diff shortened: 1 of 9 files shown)


ui/src/appComponents/GraphViewer.tsx
```diff
@@ -539,35 +496,40 @@ const GraphViewer = memo(
...
-      const [nodeSource, setNodeSource] = useState<NodeSourceResponse | null>(
-        null,
-      );
-      const [sourceLoading, setSourceLoading] = useState(false);
-      const [sourceError, setSourceError] = useState<string | null>(null);
-
-      const [nodeHistory, setNodeHistory] = useState<HistoryEntry[]>([]);
 
       // Append chat-highlighted nodes to session history
       useEffect(() => {
```
ghost, line 532: ⚠️ This ref is used to apply default filters (hiding Dependency sub-types) on the first load. However, since `GraphInteractionProvider` (the context provider) stays mounted when the user switches repositories in `App.tsx`, this ref will remain `true` and the default filters won't be applied to subsequent repositories. 

Consider resetting this flag or moving the logic to the provider and using a dependency on the graph identity.

ui/src/appComponents/GraphViewer.tsx
```diff
@@ -539,35 +496,40 @@ const GraphViewer = memo(
...
+      // Selection, filter state, history, search/hops, and derived data all
+      // live in GraphInteractionProvider so SidePanel can sit as a sibling.
+      const {
+        selectedNode,
+        setSelectedNode,
+        selectedLink,
+        setSelectedLink,
+        setNodeHistory,
+        hiddenNodeTypes,
+        setHiddenNodeTypes,
```
ghost, line 508: 🚨 **Regression**: Edge highlights (`edgeHighlightNodes`, etc.) are still managed locally in `GraphViewer` and updated only in `onLinkClick`. When an edge is selected via `SidePanel` (using `selectLink` from this context), those local highlights aren't updated. 

As a result, selecting an edge from the side panel won't highlight it on the canvas. Consider moving the edge highlight state into the provider or deriving it in `GraphViewer` from `selectedLink`.

ui/src/appComponents/GraphViewer.tsx
ghost, line 532: ⚠️ The `defaultsApplied` ref prevents the default "Dependency" filters from being applied more than once per session. However, since the `GraphInteractionProvider` stays mounted when switching repositories, this ref will remain `true`, and subsequent repositories won't have their dependency nodes hidden by default. Consider resetting this state when the repository changes.

ui/src/appComponents/SidePanel.tsx
```diff
@@ -84,28 +55,118 @@ interface SidePanelProps {
...
+  const handleSelectEdgeFromNode = useCallback(
+    (edge: NodeEdge) => {
+      if (!selectedNode) return;
+      const sourceId =
+        edge.direction === 'outgoing' ? selectedNode.id : edge.otherNodeId;
+      const targetId =
+        edge.direction === 'outgoing' ? edge.otherNodeId : selectedNode.id;
+      const nodeMap = new Map(filteredGraphData.nodes.map((n) => [n.id, n]));
+      selectLink({
+        source: sourceId,
```
ghost, line 162: ⚠️ Use the full `graphData.nodes` here instead of `filteredGraphData.nodes`. If an endpoint of the selected edge is currently hidden by filters, `nodeMap.get()` will return undefined, leading to missing node names and properties in the `EdgeDetailsPanel`.

```suggestion
      const nodeMap = new Map(graphData.nodes.map((n) => [n.id, n]));
```

ui/src/providers/GraphDataProvider.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+function InternalGraphDataProvider({ children }: { children: ReactNode }) {
+  const value = useGraphData();
+  return (
+    <GraphDataContext.Provider value={value}>
+      {children}
+    </GraphDataContext.Provider>
+  );
+}
+
```
ghost, line 56: ⚠️ The object returned by `useGraphData()` currently has a new identity on every render. Memoizing it here prevents unnecessary re-renders of the `GraphInteractionProvider` and the rest of the application tree.

```suggestion
function InternalGraphDataProvider({ children }: { children: ReactNode }) {
  const value = useGraphData();
  const memoizedValue = useMemo(() => value, [value]);
  return (
    <GraphDataContext.Provider value={memoizedValue}>
      {children}
    </GraphDataContext.Provider>
  );
}
```

ui/src/appComponents/GraphViewer.tsx
ghost: I've moved the default-hiding logic into the provider in e301c4a, where it now tracks `graphVersion`. This ensures that when you switch repositories (bumping the version), the dependency filters are correctly reapplied. Resolving this one.

ui/src/appComponents/GraphViewer.tsx
ghost: This is now handled by a dedicated `useEffect` in `GraphViewer` (e301c4a) that watches `selectedLink`. It correctly synchronizes the canvas highlights and zoom regardless of whether the selection came from a canvas click or the side panel. Closing this out.

ui/src/appComponents/GraphViewer.tsx
ghost: Handled in the provider now — see my comment on the other thread. Marking this as resolved.

ui/src/appComponents/SidePanel.tsx
ghost: Good catch. In e301c4a, I've switched `SidePanel` to use the unfiltered `graphData` for the node lookup, so edge details will now show the correct node metadata even if one side is filtered out. Resolving.

ui/src/providers/GraphDataProvider.tsx
ghost: Added the `useMemo` in e301c4a to stabilize the hook's return value. This should help prevent unnecessary re-renders in the provider tree. Closing this thread.

Edwinbhart (approved):

PR #377, opened 2026-04-29
feat(cli): headless cli

AnysSido:
## Expand and harden CLI for headless plugin integrations
🆕 **New Feature** · ✨ **Improvement** · 🐛 **Bug Fix** · ♻️ **Refactor** · 🧪 **Tests** · 🔒 **Security** · 🔧 **Chore**

This PR expands and hardens the CLI to support headless integrations like IDE plugins and MCP servers. 

It introduces repository-scoped workspaces and provides a structured JSON interface for graph search, repository metadata, and code traversal. It also refactors the indexing pipeline to ensure atomic writes and improved security during remote clones.

### Complexity
🟡 Moderate · `21 files changed, 5362 insertions(+), 146 deletions(-)`

This PR significantly expands the CLI surface area and modifies the core indexing pipeline's persistence logic. While it uses established patterns, it establishes foundational contracts (JSON shapes, workspace resolution, and exit codes) that downstream IDE plugins and MCP servers will depend on, making the consequence of design flaws or regressions moderate.

### Tests
🧪 Includes a massive test suite expansion covering git failure modes, path rehoming, and JSON contract stability across all new and modified subcommands.

### Review focus

- **Atomic indexing** — verify that the staging-to-live rename and cleanup logic in `_run_indexing_pipeline` correctly handles failure states without leaving the DB locked or corrupted.
- **Workspace resolution** — check that the `--workspace` path hashing and `_resolve_db` precedence do not break auto-discovery for existing users.
- **Security scrubbing** — confirm the regex in `_scrub_token` and the refusal of flag-like `ref` values effectively prevent credential leakage and argument injection in git operations.
- **Contract stability** — ensure the new JSON output shapes in `repos`, `source-search`, and `traverse` provide a stable and sufficient API for programmatic consumers.

### DIFF
--- a/agent/src/opentrace_agent/cli/impact.py
+++ b/agent/src/opentrace_agent/cli/impact.py
@@ -128,13 +128,14 @@ def _resolve_file_node(store: Any, file_path: str) -> dict[str, Any] | None:
         # filename input). The exact-match phase above tested
         # properties.path == <input>, which only matches files at a
         # repo root. This phase widens to any file *ending* in the
-        # basename, and requires uniqueness.
-        fts_candidates = store.search_nodes(basename, node_types=["File"], limit=20)
-        basename_matches = [
-            n
-            for n in fts_candidates
-            if _props(n).get("path", "").endswith("/" + basename) or _props(n).get("path") == basename
-        ]
+        # basename via a direct path-suffix match, and requires
+        # uniqueness. The cap is _MAX_AMBIGUOUS_CANDIDATES + 1 — one
+        # over the number the ambiguity message lists, so we can
+        # detect "more than N" without scanning further.
+        basename_matches = store.find_files_by_basename(
+            basename,
+            limit=_MAX_AMBIGUOUS_CANDIDATES + 1,
+        )
         if len(basename_matches) == 1:
             return basename_matches[0]
         if len(basename_matches) > 1:
--- a/agent/src/opentrace_agent/cli/impact.py
+++ b/agent/src/opentrace_agent/cli/impact.py
@@ -133,8 +133,7 @@ def _resolve_file_node(store: Any, file_path: str) -> dict[str, Any] | None:
         basename_matches = [
             n
             for n in fts_candidates
-            if _props(n).get("path", "").endswith("/" + basename)
-            or _props(n).get("path") == basename
+            if _props(n).get("path", "").endswith("/" + basename) or _props(n).get("path") == basename
         ]
         if len(basename_matches) == 1:
             return basename_matches[0]
@@ -367,12 +366,16 @@ def _run_json(
     """
     file_node = _resolve_file_node(store, file_path)
     if file_node is None:
-        print(json.dumps({
-            "requestedFile": file_path,
-            "file": None,
-            "symbols": [],
-            "total_dependents": 0,
-        }))
+        print(
+            json.dumps(
+                {
+                    "requestedFile": file_path,
+                    "file": None,
+                    "symbols": [],
+                    "total_dependents": 0,
+                }
+            )
+        )
         return
 
     symbols = _find_defined_symbols(store, file_node, line_ranges)
@@ -403,25 +406,32 @@ def _run_json(
         for dep in dependents:
             if dep["relationship"]["type"] in _IMPACT_RELS:
                 dep_props = _props(dep["node"])
-                sym_data["dependents"].append({
-                    "id": dep["node"]["id"],
-                    "name": dep["node"]["name"],
-                    "type": dep["node"]["type"],
-                    "relationship": dep["relationship"]["type"],
-                    "depth": dep["depth"],
-                    "path": dep_props.get("path"),
-                })
+                sym_data["dependents"].append(
+                    {
+                        "id": dep["node"]["id"],
+                        "name": dep["node"]["name"],
+                        "type": dep["node"]["type"],
+                        "relationship": dep["relationship"]["type"],
+                        "depth": dep["depth"],
+                        "path": dep_props.get("path"),
+                    }
+                )
                 total_dependents += 1
 
         result_symbols.append(sym_data)
 
     file_props = _props(file_node)
-    print(json.dumps({
-        "requestedFile": file_path,
-        "file": file_props.get("path") or file_node.get("name"),
-        "symbols": result_symbols,
-        "total_dependents": total_dependents,
-    }, default=str))
+    print(
+        json.dumps(
+            {
+                "requestedFile": file_path,
+                "file": file_props.get("path") or file_node.get("name"),
+                "symbols": result_symbols,
+                "total_dependents": total_dependents,
+            },
+            default=str,
+        )
+    )
 
 
 def _print_file_only_impact(store: Any, file_node: dict[str, Any]) -> None:
--- a/agent/src/opentrace_agent/cli/impact.py
+++ b/agent/src/opentrace_agent/cli/impact.py
@@ -15,14 +15,19 @@
 """``opentraceai impact`` — find graph nodes affected by changes to a file.
 
 Given a file path (and optional line ranges), discovers which symbols
-(functions, classes) are defined in that file and then walks incoming
-relationships (CALLS, IMPORTS, DEPENDS_ON) to surface the blast radius.
+(functions, classes) are defined in that file by walking ``File
+-DEFINES-> Symbol`` edges, then follows incoming ``CALLS``/``IMPORTS``/
+``DEPENDS_ON``/``EXTENDS``/``IMPLEMENTS`` relationships from each symbol
+to surface the blast radius.
 """
 
 from __future__ import annotations
 
+import json
 import os
 from typing import Any
+
+import click
 
 # Relationship types that indicate something *depends on* the changed symbol.
 _IMPACT_RELS = frozenset({"CALLS", "IMPORTS", "DEPENDS_ON", "EXTENDS", "IMPLEMENTS"})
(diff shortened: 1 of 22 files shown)


agent/src/opentrace_agent/cli/main.py
```diff
@@ -1084,7 +1232,693 @@ def impact(file_path: str, db_path: str | None, line_spec: str | None) -> None:
...
+
+    Uses ``splitlines()`` rather than ``split("\\n")`` so a trailing
+    newline doesn't yield a phantom empty line at the end — that would
+    inflate ``len(lines)`` by one and make open-ended ranges (``5-``)
+    print a header and blank line that don't correspond to file
+    content.
+    """
+    from pathlib import Path as P
+
+    content = P(abs_path).read_text()
```
ghost: ⚠️ `read_text()` reads the entire file into memory. For exceptionally large source files, this could cause high memory usage. Consider reading line-by-line since you're already splitting into lines anyway.

agent/src/opentrace_agent/cli/source_grep.py
```diff
@@ -0,0 +1,425 @@
...
+    # signal, etc.) is a real failure worth surfacing.
+    if proc.returncode == 1:
+        return [], False, None
+    if proc.returncode != 0:
+        msg = (proc.stderr or proc.stdout or "").strip() or f"rg exit {proc.returncode}"
+        return [], False, f"ripgrep failed: {msg}"
+
+    abs_prefix = str(repo_root) + os.sep
+    scrubbed: list[str] = []
+    for line in proc.stdout.splitlines():
```
ghost, line 228: ⚠️ `proc.stdout.splitlines()` and `scrubbed.append(...)` will collect all matches from `ripgrep` into memory before truncating to `max_count`. If `rg` finds thousands of matches across many files, this could be quite memory-intensive. Consider reading from `proc.stdout` line-by-line and stopping once `max_count` is reached.

agent/src/opentrace_agent/cli/source_grep.py
```diff
@@ -0,0 +1,425 @@
...
+        parts = rest.split(":", 2)
+        if len(parts) == 3:
+            path, line_no, text = parts
+            try:
+                parsed.append({"path": path, "line": int(line_no), "text": text})
+                continue
+            except ValueError:
+                pass
+        parsed.append({"path": None, "line": None, "text": rest})
+    return parsed
```
ghost, line 429: ⚠️ The current parsing logic will fail for filenames that contain colons (common in some environments or cross-platform paths). Since `rg`'s output format is `path:line:text`, a path like `src/parts:core.py` would result in `parts` being `['src/parts', 'core.py', '10:code']`, and `int(line_no)` would raise a `ValueError`. 

Consider using `rg --json` for a much more robust machine-readable output that avoids these parsing ambiguities.

```suggestion
        # rest is "path:line:text"
        try:
            # Split from the left but only for the first two colons to handle 
            # text containing colons. To handle paths with colons robustly, 
            # rsplit is better for the line number if we knew the format was always path:line:text.
            # However, `rg --json` is the truly robust fix.
            parts = rest.split(":", 2)
            if len(parts) == 3:
                path, line_no, text = parts
                parsed.append({"path": path, "line": int(line_no), "text": text})
                continue
        except (ValueError, IndexError):
            pass
        parsed.append({"path": None, "line": None, "text": rest})
    return parsed
```

agent/src/opentrace_agent/cli/main.py
ghost: Nice, switching to line-by-line streaming in c0a86e9 solves the memory concern for large files. I also like the cleanup of the token envvar fallbacks. Resolving this one.

agent/src/opentrace_agent/cli/source_grep.py
```diff
@@ -0,0 +1,425 @@
...
+    return up to ``N × max_count`` lines. We pass it through as a
+    safety bound on per-file work and then truncate the joined result
+    to ``max_count`` here. *truncated* is True when more matches
+    existed than were returned.
+    """
+    cmd = [
+        "rg",
+        "--no-heading",
+        "--line-number",
+        "--max-count",
```
bedonath, line 196: `rg --max-count N` is a **per-file** cap, but the surrounding code treats `max_count` as a per-repo total cap. Single-file overflows lie in the `truncated` flag.

Concrete bug: a user runs `source-grep TODO --repo myrepo --limit 50`. One file has 200 `TODO`s. ripgrep returns the first 50 from that file (and stops scanning it). The post-slice computes `truncated = (50 > 50) = False`, and emits `"truncated": false`. The IDE plugin / LLM concludes "these are all the TODOs" — but 150 were silently dropped.

The `test_truncates_total_matches_across_files` test only exercises the cross-file path (3 files × 4 matches each, limit 5), so this slips through CI.

**Fix options:**
- Drop `--max-count` from the rg argv and rely on the post-slice cap alone (the 10s timeout still protects against runaway regex), or
- pass `--max-count (max_count + 1)` so a single-file overflow trips `len(scrubbed) > max_count` honestly.

agent/src/opentrace_agent/cli/impact.py
```diff
@@ -32,10 +37,153 @@
...
+            )
+
+    basename = os.path.basename(file_path)
+    if basename:
+        # Basename fallback runs even when basename == file_path (bare
+        # filename input). The exact-match phase above tested
+        # properties.path == <input>, which only matches files at a
+        # repo root. This phase widens to any file *ending* in the
+        # basename, and requires uniqueness.
+        fts_candidates = store.search_nodes(basename, node_types=["File"], limit=20)
```
bedonath: FTS top-20 is too tight for monorepos and produces silent misses on basename-only inputs.

Concrete scenario: a frontend monorepo has 50 files named `index.ts` (one per package). User runs `opentrace impact index.ts` after editing `packages/auth/src/index.ts`.

1. Exact-match phase looks for `properties.path == "index.ts"` — only matches files at a repo root, finds nothing.
2. Basename phase: BM25 ranks all 50 candidates, returns top 20. The user's actual target may not be in the top 20 (BM25 prefers shorter/more-named paths).
3. After the `endswith("/index.ts")` filter, the count is != 1, so the function returns `None` → "file not indexed" → empty output.

The user (or the LLM consuming the hook output) sees no impact and assumes no callers exist. `impact` is the primary headless-hook use case, so a silent miss is the worst failure mode.

**Fix:** Either bump the FTS limit (e.g. 500), or — cleaner — replace this with a Cypher query that filters in the DB:

```cypher
MATCH (n:Node)
WHERE n.type = 'File'
  AND n.properties.path ENDS WITH '/' + $basename
RETURN n LIMIT N
```

That way the filter runs in the DB and isn't gated on a top-K relevance cut.

agent/src/opentrace_agent/cli/main.py
```diff
@@ -1084,7 +1232,720 @@ def impact(file_path: str, db_path: str | None, line_spec: str | None) -> None:
...
+                        break
+        except Exception:
+            pass
+
+    if not file_path:
+        # Last resort: extract from the node ID (format: repo/path/to/file.py::SymbolName)
+        if "::" in node_id:
+            candidate = node_id.split("::", 1)[0]
+            # Strip repo prefix (first segment before /)
+            parts = candidate.split("/", 1)
```
bedonath: `split("/", 1)` assumes the repo ID is a single segment with no slashes. But this same PR's `source-search` fixture and `_attribute_repo` (in `source_search.py`) explicitly support `owner/repo`-style repo IDs (`acme/widget`).

For `node_id = "acme/widget/src/parser.py::parse_widget"`:
- `candidate = "acme/widget/src/parser.py"`
- `candidate.split("/", 1) = ["acme", "widget/src/parser.py"]`
- `file_path = "widget/src/parser.py"` — wrong, should be `"src/parser.py"`

`_read_source_by_path` then searches the indexed repo paths for `widget/src/parser.py`. If nothing matches → `ClickException("Source file not found")` (loud, recoverable). But if some unrelated repo happens to have a file at that wrong relative path, the user gets back source code from the **wrong file** with no error — and the LLM consuming it reasons over that wrong content.

**Trigger:** only fires when both `properties.path` is empty AND there's no outgoing `DEFINES` File neighbor. Rare, but it's the explicit "weird cases" fallback.

**Fix:** Use the longest-prefix match against `store.list_repository_ids()` (added in this PR), same approach `_attribute_repo` in `source_search.py` already uses.

bedonath (commented):
Three correctness concerns worth addressing before merge. Inline below. Other findings (hash length, mutex timing, private `_get_neighbors` API leak, 60s git timeout, test importing `_REPOS_METADATA_FIELDS`) are fine as follow-ups.

agent/src/opentrace_agent/cli/source_grep.py
ghost: The updated logic in source_grep.py (8cd6e62) uses `max_count + 1` for the `rg` call. This ensures that if a file has more than the requested limit of matches, we'll see at least one extra result and can correctly flag the output as `truncated`. Resolving this one.

agent/src/opentrace_agent/cli/impact.py
ghost: The new `GraphStore.find_files_by_basename` helper (8cd6e62) replaces the FTS search with a direct path-suffix scan. This avoids the top-20 ranking cap and ensures we find all candidates even in large monorepos. I'll mark this as resolved.

agent/src/opentrace_agent/cli/main.py
ghost: The new `strip_repo_prefix` helper (8cd6e62) correctly handles multi-segment repository IDs by matching against the list of indexed repo IDs (longest-first). This fixes the over-stripping bug for `owner/repo` style IDs. Resolving.

bedonath (approved):

PR #382, opened 2026-05-01
feat(plugin): opencode plugin

AnysSido:
## OpenCode plugin and multi-repo indexing support
🆕 **New Feature** · ✨ **Improvement** · 🐛 **Bug Fix** · 🧪 **Tests** · 📝 **Docs** · 🔧 **Chore** · 🔒 **Security**

This PR introduces the `@opentrace/opencode` plugin, enabling deep codebase intelligence within the OpenCode AI agent. 

It also upgrades the OpenTrace agent to support multi-repo workspaces via incremental indexing, atomic database swaps, and cross-process file locking.

### Key Changes
- **OpenCode Plugin**: Adds 9 tools for search and graph traversal, along with hooks that automatically augment standard search/edit tools with graph context.
- **Multi-repo Indexing**: Refactors the agent's indexing pipeline to append new repositories to existing workspaces instead of replacing them.
- **Atomic Persistence**: Implements a "seed-and-swap" strategy for the database to ensure indexing is robust against crashes or interruptions.
- **Concurrency Control**: Adds POSIX file locking to prevent race conditions when multiple agents or CLI instances attempt to index the same workspace.
- **Auth Integration**: Implements secure PAT resolution for GitHub and GitLab with built-in redaction for debug logs.

### Complexity
🟠 High · `65 files changed, 7387 insertions(+), 64 deletions(-)`

This PR spans multiple system layers, introducing a substantial TypeScript plugin while simultaneously modifying core Python agent logic for workspace management. 

The changes to the indexing pipeline—specifically the introduction of file locking and the atomic database swap—are foundational and carry high consequence for data integrity. The plugin's architecture sets the pattern for how OpenTrace integrates with AI runtimes through both tools and automated context-augmentation hooks.

### Tests
🧪 Includes a comprehensive suite of 240+ unit tests for the plugin and new Python regression tests for multi-repo indexing and locking.

### Review focus

- **Atomic DB Swap** — verify the `_swap_staging_into_place` logic in `main.py` correctly handles the DB/WAL pair to prevent corruption.
- **File Locking** — ensure the `fcntl.flock` implementation provides a clean UX for concurrent indexing without leaving stale lockfiles.
- **Auth Leakage** — confirm that the `redactArgs` helper and hostname matching in `auth.ts` sufficiently protect git credentials.
- **Hook Performance** — check that the 3-second timeouts on system-prompt and tool hooks are sufficient to prevent UI lag.


plugins/opencode/src/util/node-id.ts
```diff
@@ -0,0 +1,64 @@
...
+  const symbol = doubleColon !== -1 ? nodeId.substring(doubleColon + 2) : null
+  const slashIdx = pathPart.indexOf("/")
+  if (slashIdx === -1) {
+    return { repo: pathPart, path: null, symbol }
+  }
+  return {
+    repo: pathPart.substring(0, slashIdx),
+    path: pathPart.substring(slashIdx + 1),
+    symbol,
+  }
```
ghost, line 47: ⚠️ If `repoId` can contain slashes (e.g., `owner/repo`), this simple split will incorrectly identify the repo. The agent uses a more sophisticated approach by matching against known repo IDs. If we can't do that here, we should at least consider if `repoId` is guaranteed to be a single segment.

```suggestion
  const slashIdx = pathPart.indexOf("/")
  if (slashIdx === -1) {
    return { repo: pathPart, path: null, symbol }
  }
  // Note: This assumes repoId has no slashes. If OpenTrace supports 
  // "org/repo" style IDs, this logic will need to match against 
  // a known list of repo IDs similar to the agent's implementation.
  return {
    repo: pathPart.substring(0, slashIdx),
    path: pathPart.substring(slashIdx + 1),
    symbol,
  }
```

ghost: Just a heads-up that this parsing logic is still in place as of 960834a. Since the PR description mentions multi-repo support, we'll likely hit issues if `repoId` contains slashes (which is common for GitHub/GitLab). 

Adding a comment acknowledging the assumption or matching against a known list of repo IDs (like the agent does) would make this more robust.

Edwinbhart (approved):

PR #394, opened 2026-05-08
feat(wiki): an initial implementation of vaults

callumyuille:
## Knowledge vaults v2 and unified graph retrieval primitives
🆕 **New Feature** · ✨ **Improvement** · ♻️ **Refactor** · 🧪 **Tests** · 📝 **Docs**

Introduces Knowledge Vaults v2 with a multi-stage LLM pipeline for ingesting folder-based content into the graph.
Consolidates graph querying into 8 unified retrieval primitives consumed by MCP, the REST API, and the UI.

### Complexity
🟠 High · `98 files changed, 13139 insertions(+), 377 deletions(-)`

This is a broad-scope PR that introduces a new architectural layer for knowledge ingestion and retrieval. It touches every part of the system: Proto definitions, core agent logic, CLI commands, REST/MCP interfaces, and the React-based UI. The changes establish foundational patterns for how agents interact with external documentation and how that data is mirrored into the graph.

### Tests
🧪 Includes an extensive new test suite covering the full compilation pipeline, retrieval primitives, and MCP integration.

### Review focus

- **LLM structured output** — verify the `WikiLLM` implementation in `wiki/llm.py` correctly handles tool-calling and retries across Anthropic, Gemini, and OpenAI.
- **Graph mirroring** — check `wiki/ingest/graph_writer.py` for correct handling of `WikiVault` and `WikiPage` node/edge upserts in the graph store.
- **Concurrency** — review the `_flock` implementation in `pipeline.py` to ensure vault metadata is protected during concurrent compile attempts.


agent/src/opentrace_agent/store/graph_store.py
```diff
@@ -263,6 +276,36 @@ def merge_relationship(
...
+    def delete_node(self, node_id: str) -> bool:
+        """Delete a node and any edges touching it (DETACH DELETE).
+
+        Returns ``True`` when a node existed and was removed, ``False`` when
+        no such node was present.
+        """
+        if self.get_node(node_id) is None:
+            return False
+        # Detach incoming and outgoing edges first — LadybugDB DETACH DELETE
+        # support varies, so we delete edges by RELATES match then the node.
```
ghost, line 288: 🚨 **Schema Mismatch between Python and TS implementations.**

This implementation (and others in this file like `traverse`) still hardcodes the single `Node` table:
```python
"MATCH (a:Node {id: $id})-[r:RELATES]->(:Node) DELETE r"
```
However, the TypeScript `LadybugGraphStore` (used in browser-mode) has been migrated to typed tables (`Repository`, `File`, etc.) using `NODE_SCHEMA_STATEMENTS` from `schema_gen.py`.

This breaks direct file-level interoperability of LadybugDB files between the CLI/Agent and the Browser. A database file indexed on the CLI will not be readable by the browser in in-memory mode, and vice versa.

If this is a deliberate phased rollout, we should at least document it. If not, the Python implementation should also be updated to use the typed schema.

agent/src/opentrace_agent/cli/serve.py
```diff
@@ -193,3 +379,180 @@ async def health(request: Request) -> JSONResponse:
...
+
+        from opentrace_agent.wiki.paths import InvalidVaultName, pages_dir
+
+        name = request.path_params["vault"]
+        slug = request.path_params["slug"]
+        # Slug must look like a slug — no separators.
+        if "/" in slug or ".." in slug or slug.startswith("."):
+            return _error(400, f"invalid slug: {slug}")
+        try:
+            pd = pages_dir(name)
```
ghost, line 438: ⚠️ **Potential memory issue with large file uploads.**

The `compile_route` reads all uploaded files into memory simultaneously:
```python
data = await f.read()
```
If a user uploads many large files (e.g. documentation PDFs or large source code repositories), this could lead to high memory consumption or OOM failures on the server. Consider streaming the files to disk or processing them one by one if possible.

ui/src/components/wiki/VaultBrowser.tsx
```diff
@@ -0,0 +1,369 @@
...
+      setActiveVault(name);
+      const data = await getVault(name);
+      setVaultData(data);
+      if (data.pages.length > 0) setActiveSlug(data.pages[0].slug);
+      // Reload the graph so the newly-mirrored vault/page/source nodes
+      // appear without a hard refresh. Server-mode is read-through; in
+      // browser-mode this is a no-op against the in-browser store.
+      void loadGraph();
+    },
+    [refreshVaults, loadGraph],
```
ghost, line 143: ⚠️ **Fragmented experience between Browser and Server modes.**

The "Vaults" feature is exclusively server-backed (it always hits the API and expects the server to manage disk/graph storage), while the Graph can be either Browser-backed (in-memory) or Server-backed.

In Browser-mode, newly compiled vaults will not appear in the graph because the server updates its own database, not the browser's in-memory one. We might want to suggest the user switch to "Server mode" or provide a way to sync the browser's DB after a vault is added.

agent/src/opentrace_agent/cli/mcp_server.py
```diff
@@ -130,41 +192,224 @@ def traverse_graph(
...
+        edgeTypes: str = "",
+        vaultScope: str = "",
+        confidenceThreshold: float = 0.0,
     ) -> str:
         """Walk relationships from a starting node.
 
         Direction can be 'outgoing', 'incoming', or 'both'.
-        Optionally filter by relationship type (e.g. 'CALLS', 'DEFINES', 'CONTAINS').
+
+        Filter the walk by relationship type via either ``relationship`` (a
```
ghost, line 217: ⚠️ **Docstring is outdated.**

The docstring says `vaultScope` and `confidenceThreshold` are currently no-ops, but they are actually implemented in `GraphStore.traverse` and passed through here.

```suggestion
        Filter the walk by relationship type via either ``relationship`` (a
        single type, e.g. 'CALLS') or ``edgeTypes`` (a comma-separated set,
        e.g. 'CALLS,IMPORTS'). When both are given, ``edgeTypes`` wins.

        Optional ``vaultScope`` restricts traversal to nodes in a specific
        vault; ``confidenceThreshold`` (0.0-1.0) skips low-confidence edges.
        """
```

ghost: In a5902689, you've implemented the logic for `vault_scope` and `confidence_threshold` in `GraphStore.traverse`, which is great. However, the docstrings in both `graph_store.py` and `mcp_server.py` still describe them as "currently no-op". Might want to clean those up to reflect the actual implementation.

ui/src/components/wiki/VaultBrowser.tsx
ghost: The new `wiki backfill` command and the updated instructions in `README.md` (added in a5902689) regarding the `?server=` query parameter go a long way toward bridging the gap between browser and server modes. It's much clearer now how to keep things in sync. Resolving this one.

ghost:
## 🔄 Review Update
> 1 new commit: a590268

The heavy lifting on Knowledge Vaults v2 is impressive—consolidating retrieval into those 8 primitives makes the architectural layer much cleaner.

I've resolved the thread about the fragmented browser/server experience; the new `backfill` command and the updated setup docs bridge that gap nicely.

A few things are still outstanding:
- The [schema mismatch](https://github.com/opentrace/opentrace/pull/394#discussion_r3208516747) remains. Now that the typed tables are defined in `schema_gen.py`, it would be great to see the Python store migrate away from the hardcoded `Node` table to ensure full interoperability.
- The memory concerns with large file uploads in `compile_route` ([thread](https://github.com/opentrace/opentrace/pull/394#discussion_r3208516754)) haven't been tackled yet.
- I noticed you've implemented the logic for `vault_scope` and `confidence_threshold` in `traverse`, but the docstrings in both the store and the MCP server still label them as no-ops.

This is a huge step forward for the wiki feature. Once the schema and memory items are addressed, it should be ready to go.

agent/src/opentrace_agent/cli/serve.py
callumyuille: We can address this in later iterations

agent/src/opentrace_agent/store/graph_store.py
callumyuille: The divergence is long-standing rather than a fresh design call — Python's single-`Node` schema predates the typed tables, and we documented it on this branch (in `agent/src/opentrace_agent/store/CLAUDE.md`) when we wired the 
  proto-generated type list into Python via `VALID_NODE_TYPES`:                                                                                                                                                                     
                                                                                                                                                                                                                                    
  > The Python `GraphStore` here ignores those typed tables and creates its own simpler schema… Both stores accept the same logical node types and relationship types; the storage shape differs.                                   
                                                                      
  So the two stores share the *logical* type system (sourced from `gen/schema_gen.py` in both) but not the storage shape. Migrating Python to typed tables would mean rewriting every Cypher query in `graph_store.py` to be        
  per-label, and the consumers above (REST/MCP) wouldn't notice — they only see logical types.
                                                                                                                                                                                                                                    
  For the cross-mode interop concern specifically: file-level interop has never been the contract — the contract is REST/MCP. The new `opentraceai wiki backfill` command + the `?server=` URL param give users a path to share     
  state between modes without needing a binary-compatible DB file.
                                                                                                                                                                                                                                    
  Worth a follow-up ticket if we decide file portability is a goal we want to honour

agent/src/opentrace_agent/cli/mcp_server.py
ghost: I see you've updated the docstrings in 3092468a to accurately describe the `vault_scope` and `confidence_threshold` parameters. This makes the implementation much clearer. I'll mark this as resolved.

ghost:
## ✅ Review Update — Approved
> 1 new commit: 3092468

The docstring updates for the retrieval parameters in the store and MCP server look perfect—much better than having "no-op" placeholders in there when the logic is actually live.

I've resolved the documentation thread. Since we've agreed to handle the schema alignment and memory streaming in future iterations, this is looking solid and ready to land.

brandontaslater (approved):

brandontaslater:
@callumyuille before you merge, have you tested this library update from within the cloud UI project?

brandontaslater (changes requested):
@callumyuille before you merge, have you tested this library update from within the cloud UI project?

callumyuille:
> @callumyuille before you merge, have you tested this library update from within the cloud UI project?

No cloud-UI testing was needed — this branch doesn't update any library that insight-ui consumes.
The only npm change is remark-wiki-link@^2.0.1 added to opentrace's ui/package.json (a separate React app), not to insight-ui. The proto changes are additive (new wiki node/edge types + reserved JobPhase slots that aren't emitted yet), so even if insight-ui regenerates from the shared .proto, there's no behavior change to test — protobuf tolerates unknown fields and no new enum values are sent on the wire.

brandontaslater (approved):

PR #395, opened 2026-05-11
fix(agent): resolve TS named-import bindings for cross-file CALLS

jatinot:
## Fix TypeScript named-import resolution for cross-file calls
🐛 **Bug Fix** · 🧪 **Tests** · 🔧 **Chore**

Fixes a bug where TypeScript named imports (e.g., `import { foo } from './utils'`) failed to resolve bare call sites in the graph.

Correctly extracts all local bindings—including named, default, aliased, and namespace imports—using tree-sitter.

Preserves original path-basename aliases as a fallback for namespace-style attribute calls.

### Complexity
🟢 Low · `4 files changed, 161 insertions(+), 4 deletions(-)`

Targeted fix within the TypeScript import analyzer and corresponding test updates. It maintains existing behavior as a fallback to ensure no regressions in path-based resolution.

### Tests
🧪 Includes regression tests for the full pipeline and comprehensive unit tests for various TypeScript import syntaxes.

### Review focus

- **Tree-sitter traversal** — ensure the fallback logic for identifier extraction correctly captures the intended binding name across different grammar versions.


agent/src/opentrace_agent/sources/code/import_analyzer.py
```diff
@@ -390,14 +390,79 @@ def _parse_ts_import(
...
+            elif sub.type == "namespace_import":
+                # `* as ns`
+                for ns_child in sub.children:
+                    if ns_child.type == "identifier":
+                        bindings.append(ns_child.text.decode())
+                        break
+    return bindings
+
 
 def _parse_ts_reexport(
```
ghost, line 467: ⚠️ Should `_parse_ts_reexport` also be updated to map specific bindings? Currently it still uses the path-basename alias logic, which might cause similar resolution issues for re-exported symbols.

jatinot (commented):
I now have everything I need. Here is the complete review:

---

## Summary

PR #395 fixes a real and impactful bug: `_parse_ts_import` previously registered only the path basename (e.g. `utils`) as the import alias, so bare calls to named imports (e.g. `foo()` from `import { foo } from './utils'`) were silently dropped from the CALLS graph. The fix introduces a new helper `_ts_import_bindings` that walks the tree-sitter AST to extract all local bindings (default, named, namespaced, aliased) and registers each one. The approach is architecturally sound, well-commented, and well-tested. There are, however, several issues worth addressing — one correctness bug and several quality concerns.

---

## Issues Found

### **[Bugs & Logic Error] Severity: High** — `test_import_analyzer.py:174–179` (base file) / `test_import_analyzer.py` after the diff

**`test_relative_import` is now a stale, contradictory test that will likely fail after the fix.**

The base (pre-PR) `test_relative_import` test at line 174 of the test file tests `import { helper } from './utils'` and asserts `result.internal.get("utils") == "src/utils.ts"` — checking that the basename alias is stored. The PR adds a new test `test_named_import_binding` which uses the **same source input** and asserts that `"helper"` (the named binding) is stored. 

Looking at the new code, the PR **renames** `test_relative_import` to `test_named_import_binding` and changes its assertion from `"utils"` to `"helper"`. However, the **base file** (which the new test file content replaces) still shows the original `test_relative_import` body at line 174 with `assert result.internal.get("utils") == "src/utils.ts"`. In the PR diff, this test is **replaced** but the test `test_relative_import_parent_dir` still only asserts the basename `"config"`, not the named binding `"config"` (they happen to coincide in that case). The concern is there's no test asserting the **co-existence** of both `"helper"` (new binding) and `"utils"` (basename fallback) in the same call — which the new code now produces for `import { helper } from './utils'`.

**Suggested fix:** Add an assertion to `test_named_import_binding` that also verifies `result.internal.get("utils") == "src/utils.ts"` (the basename alias should still be registered), confirming the backward-compatible fallback works.

---

### **[Bugs & Logic Error] Severity: Medium** — `import_analyzer.py`, `_ts_import_bindings`, fallback path

**The fallback for grammars not exposing field names has an unreachable `continue`.**

```python
alias_node = spec.child_by_field_name("alias")
name_node = spec.child_by_field_name("name")
bound = alias_node or name_node
if bound is not None:
    bindings.append(bound.text.decode())
    continue          # ← jumps back to `for spec in sub.children:`
# Fallback for grammars that don't expose field names:
idents = [c for c in spec.children if c.type == "identifier"]
if idents:
    bindings.append(idents[-1].text.decode())
```

When `bound is not None` (the field-name strategy succeeds), the `continue` correctly skips the fallback — that's fine. But the **comment** above the fallback says "Fallback for grammars that don't expose field names", implying this is meant to be a second-chance path for when *both* `alias_node` and `name_node` are `None`. That logic is correct. However, there's a subtle issue: if `child_by_field_name` is available but returns `None` for both `"alias"` and `"name"` *only in degenerate/malformed specifier nodes*, the fallback code runs but no `continue` prevents double-appending if the fallback path is later extended. This is low-risk as written, but the structure can be made clearer by using an `else` branch instead.

**Suggested fix:**
```python
bound = alias_node or name_node
if bound is not None:
    bindings.append(bound.text.decode())
else:
    # Fallback for grammars that don't expose field names
    idents = [c for c in spec.children if c.type == "identifier"]
    if idents:
        bindings.append(idents[-1].text.decode())
```

---

### **[Design & Architecture] Severity: Medium** — `import_analyzer.py`, `_parse_ts_reexport` not updated

**`_parse_ts_reexport` has the same basename-only alias bug that was just fixed in `_parse_ts_import`, and this PR doesn't address it.**

```python
def _parse_ts_reexport(...) -> None:
    ...
    for ext in extensions:
        candidate = resolved + ext
        if candidate in known_files:
            alias = source_text.rsplit("/", 1)[-1]   # ← still uses basename only
            result[alias] = candidate
            break
```

The UI-side counterpart `parseTsReexport` in `importAnalyzer.ts` (line 438–476) has the same basename-only implementation, and neither has been updated. `export { Config } from './config'` would fail to resolve a bare `Config()` call for the same reason as the original bug. This is an explicit parallel codepath that should receive the same treatment — either fix it in this PR (since it lives in the same function file) or file a follow-up issue.

---

### **[Design & Architecture] Severity: Medium** — UI-side `parseTsImport` not ported

**The fix is applied only to the Python agent but not to the equivalent TypeScript UI implementation.**

`ui/src/components/pipeline/parser/importAnalyzer.ts::parseTsImport` (line 396–434) uses the exact same old basename-only strategy:
```typescript
const alias = sourceText.split('/').pop()!;
internal[alias] = candidate;
```

The file header says *"Ported from agent/src/opentrace_agent/sources/code/import_analyzer.py"*, making clear these two are expected to stay in sync. The UI pipeline will continue to miss cross-file CALLS edges for named imports, leaving the fix incomplete for browser-side indexing. This should either be addressed in this PR or tracked explicitly.

---

### **[Naming & Clarity] Severity: Low** — `test_import_analyzer.py:174`, test name no longer matches intent

The original `test_relative_import` was renamed to `test_named_import_binding` in the diff, which is correct. However, the **residual** `test_relative_import` in the base file (which in the new file becomes the renamed test) shares its source code and `known` set with the new `test_named_import_binding`. It now only checks `"utils"` via the *basename fallback*, but there's no test in the class that's simply named `test_relative_import` that verifies the basic resolution works end-to-end. The renamed test should perhaps be split: one for named-binding (`test_named_import_binding`) and one for path resolution (`test_relative_import`), keeping both concerns separate and making regressions easier to diagnose.

---

### **[Bugs & Logic Error] Severity: Low** — `_ts_import_bindings`: `import type` bindings are extracted unnecessarily

The docstring lists `import type { Foo } from '...'` → `["Foo"]` as a supported case. Type-only imports (`import type`) introduce no runtime value binding and should never produce a CALLS edge. Storing them in the `result` map is harmless today (a type can't be called), but it pollutes the alias registry and could cause false positives if a value and a type share the same name.

**Suggested fix:** Check the import statement node for a `type` keyword before calling `_ts_import_bindings`, and skip or filter type-only imports.

---

### **[Error Handling] Severity: Low** — `source_text` can equal `"."` or `".."` resulting in an empty `basename_alias`

```python
basename_alias = source_text.rsplit("/", 1)[-1]
if basename_alias and basename_alias not in result:
    result[basename_alias] = candidate_path
```

The `if basename_alias` guard protects against an empty string but not against `"."` or `".."` being written as an alias (from a malformed import like `import foo from '.'`). In practice `_resolve_relative_path` handles this, but `source_text.rsplit("/", 1)[-1]` on `"."` produces `"."` (a one-character string, truthy), which would enter the registry as the key `"."`. This is an edge case but defensive filtering against non-identifier characters would be cleaner.

---

## Positive Observations

- **Well-identified root cause.** The commit message and inline comments accurately describe why the old approach (`rsplit("/", 1)[-1]`) was fundamentally wrong, making the change easy to audit.
- **Comprehensive grammar coverage.** `_ts_import_bindings` handles default, named, aliased, namespace, and type-only imports in a single clear loop, with explicit comments for each case. The two-tier strategy (field names first, identifier fallback) is pragmatic and handles real-world grammar variations.
- **Good test hygiene.** The unit tests in `test_import_analyzer.py` are granular and well-named; each import form has its own test. The regression test in `test_pipeline.py` is particularly valuable — it's a true end-to-end verification that a CALLS edge actually appears in the output graph.
- **Backward-compatible basename fallback preserved.** The PR is careful not to break the existing namespace-style attribute call resolution (`utils.helper()`) by keeping the basename alias as a secondary registration, with correct priority to explicit bindings.
- **`.gitignore` change is correct.** Adding `*.indexlock` prevents SQLite lock files from polluting the index snapshot — a sensible housekeeping fix.

---

```json:review
{
  "summary": "This PR correctly identifies and fixes a real bug where `_parse_ts_import` was registering only the path basename as the import alias, causing CALLS edges for named TypeScript imports to silently disappear. The new `_ts_import_bindings` helper is well-structured and well-tested. However, there are several issues: (1) the parallel `_parse_ts_reexport` function has the exact same basename-only bug and is not fixed; (2) the UI-side `parseTsImport` in `importAnalyzer.ts` is explicitly documented as a port of the Python code but is not updated, leaving the UI pipeline broken for the same case; (3) the fallback path in `_ts_import_bindings` uses a `continue` statement where a cleaner `else` block would better express the intent; (4) `import type` bindings (type-only, never callable) are extracted into the alias registry unnecessarily; (5) the base `test_relative_import` test that this PR effectively replaces was checking basename-only resolution — the new tests don't confirm the co-existence of both the named binding *and* the basename fallback for the same import statement. The core logic is sound and the regression tests are a meaningful improvement; the architectural gaps (reexport path, UI sync) should be addressed before merging.",
  "verdict": "REQUEST_CHANGES",
  "comments": [
    {
      "path": "agent/src/opentrace_agent/sources/code/import_analyzer.py",
      "line": 430,
      "body": "**`_parse_ts_reexport` has the same basename-only alias bug that this PR fixes in `_parse_ts_import`.** The `export { Config } from './config'` path still uses `alias = source_text.rsplit('/', 1)[-1]`, registering only the path basename rather than the exported binding names. A bare `Config()` call in the importing file would fail to resolve for the same reason as the original bug. Since `_parse_ts_reexport` lives in the same file, it should receive the same treatment — either apply `_ts_import_bindings` (or a reexport-equivalent helper) here, or file a tracked follow-up issue."
    },
    {
      "path": "agent/src/opentrace_agent/sources/code/import_analyzer.py",
      "line": 463,
      "body": "**Prefer `else` over `continue` for the fallback branch.** The current structure (`if bound is not None: ...; continue; # fallback`) is slightly misleading — the `continue` makes the fallback block look like it's inside a condition that's separate from the `if bound is not None` block, but structurally it's an implicit `else`. Using `else` directly expresses this intent and prevents potential double-appending if the block is ever extended:\n```python\nif bound is not None:\n    bindings.append(bound.text.decode())\nelse:\n    # Fallback for grammars that don't expose field names\n    idents = [c for c in spec.children if c.type == \"identifier\"]\n    if idents:\n        bindings.append(idents[-1].text.decode())\n```"
    },
    {
      "path": "agent/src/opentrace_agent/sources/code/import_analyzer.py",
      "line": 420,
      "body": "**`import type { Foo }` bindings are extracted into the alias registry.** The docstring documents `import type { Foo }` → `[\"Foo\"]` as a supported case, but type-only imports carry no runtime value and cannot be called. Storing them in `result` pollutes the alias map and could cause false-positive CALLS edges if a type and a value share the same name. Consider detecting `import type` at the call site in `_parse_ts_import` (before calling `_ts_import_bindings`) and returning early, or filtering the output of `_ts_import_bindings`."
    },
    {
      "path": "agent/tests/opentrace_agent/sources/code/test_import_analyzer.py",
      "line": 181,
      "body": "**`test_named_import_binding` should also assert the basename fallback is registered.** The new code registers *both* `\"helper\"` (named binding) and `\"utils\"` (basename fallback) for `import { helper } from './utils'`. The test only checks `\"helper\"`, leaving the backward-compatible fallback untested. Add: `assert result.internal.get(\"utils\") == \"src/utils.ts\"` to confirm both keys are present. Also, the pre-existing `test_relative_import` (line 174 in the base file) still asserts only `\"utils\"` — confirm it has been fully replaced or updated to reflect the new behaviour."
    },
    {
      "path": "agent/src/opentrace_agent/sources/code/import_analyzer.py",
      "line": 413,
      "body": "**`basename_alias` can be `\".\"` or `\"..\"` for edge-case imports like `import foo from '.'`.** The guard `if basename_alias and basename_alias not in result` correctly filters empty strings but passes through `\".\"` and `\"..\"` (both truthy). These would be written into `result` as invalid alias keys. Consider adding a check: `if basename_alias and not basename_alias.startswith('.') and basename_alias not in result:`."
    },
    {
      "body": "**The UI-side `parseTsImport` in `ui/src/components/pipeline/parser/importAnalyzer.ts` (line 396–434) is not updated.** The file header explicitly says it is \"ported from\" the Python agent implementation. It uses the same basename-only strategy (`const alias = sourceText.split('/').pop()!`) that this PR fixes in Python. The browser-side indexing pipeline will continue to drop CALLS edges for named TypeScript imports. This should either be fixed in this PR or tracked as a follow-up issue, given the explicit parity expectation between the two files."
    }
  ]
}
```

---
*Generated via [OpenTrace](https://app.opentrace.ai)*

PR #397, opened 2026-05-12
fix(plugins/opencode): support Node runtime for OpenCode Desktop

brandontaslater:
## Support Node.js runtime for OpenCode plugin
✨ **Improvement** · 🐛 **Bug Fix** · ♻️ **Refactor** · 🧪 **Tests** · 🔧 **Chore**

This PR migrates the OpenCode plugin to use cross-runtime helpers instead of Bun-specific APIs. 

This fixes a failure where the plugin would crash when loaded by the Node-based Electron sidecar in OpenCode Desktop. It ensures the same bundle works across both Bun (CLI/TUI) and Node.js (Desktop).

### Complexity
🟡 Moderate · `7 files changed, 374 insertions(+), 49 deletions(-)`

The PR replaces foundational runtime APIs with custom implementations for process management and file discovery. While the scope is limited to a single plugin, the change involves manual handling of streams, process lifecycle events, and cross-platform path resolution, which carry subtle concurrency and portability risks.

### Tests
🧪 Includes 12 new unit tests for the cross-runtime process and PATH helpers, covering pipe overflows, timeouts, and edge-case discovery.

### Review focus

- **Pipe drainage** — verify that `runCommand` correctly drains stdout/stderr via events to prevent deadlocks on chatty subprocesses.
- **Process exit semantics** — confirm that resolving on the `exit` event (rather than `close`) is the correct choice for avoiding hangs when grandchildren inherit pipes.
- **Windows PATH resolution** — ensure `whichSync` correctly handles `PATHEXT` and directory skipping to match standard executable discovery.

### DIFF
--- a/plugins/opencode/src/util/process.ts
+++ b/plugins/opencode/src/util/process.ts
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2026 OpenTrace Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Cross-runtime process + PATH helpers.
+ *
+ * Both APIs route through `node:child_process` and `node:fs`, which Bun
+ * fully reimplements. That means the same bundle works under Bun (CLI /
+ * TUI) and Node (the OpenCode desktop's Electron utility process). The
+ * earlier `Bun.spawn` / `Bun.which` path threw a synchronous
+ * ReferenceError in the desktop sidecar, causing the plugin's `server()`
+ * factory to reject after `configureDebug()` and silently abandon every
+ * hook registration. See git blame for the diagnosis trail.
+ */
+
+import { spawn } from "node:child_process"
+import { statSync } from "node:fs"
+import { delimiter, join } from "node:path"
+
+export interface RunResult {
+  /** Process exit code. -1 when killed before exit (e.g. timeout, signal). */
+  exitCode: number
+  stdout: string
+  stderr: string
+  /** True when the kill timer fired before the process exited on its own. */
+  timedOut: boolean
+}
+
+/**
+ * Spawn a process, accumulate stdout/stderr, and resolve on exit. Mirrors
+ * the shape we previously got from `Bun.spawn` + `proc.exited` + a manual
+ * `new Response(proc.stdout).text()`, in a single call.
+ *
+ * Stdout/stderr are drained via `data` events so the pipes never deadlock
+ * on a chatty subprocess — the prior Bun-based code intentionally read
+ * the streams before awaiting exit for the same reason.
+ */
+export async function runCommand(
+  cmd: string,
+  args: string[],
+  opts: { env?: NodeJS.ProcessEnv; timeoutMs?: number } = {},
+): Promise<RunResult> {
+  return new Promise<RunResult>((resolve, reject) => {
+    let child
+    try {
+      child = spawn(cmd, args, {
+        env: opts.env ?? process.env,
+        stdio: ["ignore", "pipe", "pipe"],
+      })
+    } catch (e) {
+      // Synchronous spawn errors are rare (Node usually reports via the
+      // async "error" event), but reject so callers can match on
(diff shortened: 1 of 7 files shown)


plugins/opencode/src/util/process.ts
```diff
@@ -0,0 +1,152 @@
...
+        stderr: Buffer.concat(stderrChunks).toString("utf8"),
+        timedOut,
+      })
+    })
+  })
+}
+
+/**
+ * PATH-scan replacement for `Bun.which`. Returns the absolute path of the
+ * first executable named *name* found on *PATH*, or null. On Windows,
```
ghost, line 116: ⚠️ Resolving on `exit` and immediately calling `destroy()` can truncate output because stdio pipes are asynchronous in Node.js. Even if the parent process has exited, its output may still be in flight or buffered in Node's stream. 

The previous Bun-based code used `Promise.all([proc.exited, stdoutPromise, stderrPromise])`, which waited for the stdout stream to reach EOF. 

The safest way to match this behavior in Node and ensure all data is captured is to use the `close` event, which only fires after all stdio pipes have been closed. If you are worried about grandchildren holding the pipes open indefinitely, it's better to wait for `close` but have a separate fallback timeout that calls `destroy()` if it hasn't closed shortly after the process exited.

However, given this was working with `Promise.all` in Bun (which also waits for EOF), using `close` should be fine.

```suggestion
    child.once("error", (e) => {
      if (killTimer) clearTimeout(killTimer)
      reject(e)
    })
    child.once("close", (code, signal) => {
      if (killTimer) clearTimeout(killTimer)
      resolve({
        exitCode: code ?? -1,
        stdout: Buffer.concat(stdoutChunks).toString("utf8"),
        stderr: Buffer.concat(stderrChunks).toString("utf8"),
        timedOut,
      })
    })
  })
}
```

plugins/opencode/src/util/process.ts
```diff
@@ -0,0 +1,152 @@
...
+      try {
+        const st = statSync(candidate)
+        if (st.isFile()) return candidate
+      } catch {
+        // ENOENT / ENOTDIR / EACCES — keep scanning.
+      }
+    }
+  }
+  return null
+}
```
ghost, line 141: ⚠️ On Unix systems, `which` should verify that the file is actually executable. A file might exist with the same name (e.g., a directory or a non-executable config file) but it cannot be run as a command. `Bun.which` performs this check.

```suggestion
import { accessSync, constants, statSync } from "node:fs"
import { delimiter, join } from "node:path"

export interface RunResult {
// ...
// (I will only replace the whichSync part in the actual call)

```

plugins/opencode/src/util/process.ts
```diff
@@ -0,0 +1,152 @@
...
+
+function splitPathExt(pathext: string | undefined): string[] | null {
```
ghost, line 143: ⚠️ On Unix systems, `which` should verify that the file is actually executable by the current user. `Bun.which` performs this check; without it, `whichSync` might return a file that exists but cannot be spawned, leading to downstream errors.

```suggestion
    for (const ext of exts) {
      const candidate = join(dir, name + ext)
      try {
        const st = statSync(candidate)
        if (st.isFile()) {
          if (process.platform === "win32") return candidate
          // Check if executable by the current user
          accessSync(candidate, constants.X_OK)
          return candidate
        }
      } catch {
        // ENOENT / ENOTDIR / EACCES / etc. — keep scanning.
      }
    }
```

plugins/opencode/src/util/process.ts
```diff
@@ -0,0 +1,152 @@
...
+ * fully reimplements. That means the same bundle works under Bun (CLI /
+ * TUI) and Node (the OpenCode desktop's Electron utility process). The
+ * earlier `Bun.spawn` / `Bun.which` path threw a synchronous
+ * ReferenceError in the desktop sidecar, causing the plugin's `server()`
+ * factory to reject after `configureDebug()` and silently abandon every
+ * hook registration. See git blame for the diagnosis trail.
+ */
+
+import { spawn } from "node:child_process"
+import { statSync } from "node:fs"
```
ghost, line 30: Add necessary imports for executability check.

```suggestion
import { spawn } from "node:child_process"
import { accessSync, constants, statSync } from "node:fs"
import { delimiter, join } from "node:path"
```

plugins/opencode/src/util/process.ts
```diff
@@ -0,0 +1,152 @@
...
+        // Killed-by-signal exits report a null code; surface -1 so callers
+        // can distinguish from a normal "exit 0".
+        exitCode: code ?? (signal ? -1 : -1),
+        stdout: Buffer.concat(stdoutChunks).toString("utf8"),
+        stderr: Buffer.concat(stderrChunks).toString("utf8"),
+        timedOut,
+      })
+    })
+  })
+}
```
ghost, line 112: Can be simplified to `code ?? -1`. Node's `exit` event always provides either a code or a signal.

```suggestion
        exitCode: code ?? -1,
```

PR #398, opened 2026-06-08
feat: Claude Code plugin parity — bundled MCP server, skills, hooks, and agent tooling

jatinot:
## Summary

Brings the Claude Code plugin to parity with (and past) the OpenCode plugin. The plugin now ships a bundled stdio MCP server (`uvx opentraceai mcp`, 12 tools), a skill-based routing layer, a full hook suite, and the agent-side MCP tooling those skills depend on.

## Changes

**Claude Code plugin**
- Restructure around skill-based routing: 11 `opentrace-*` skills (explore, interrogate, find-usages, impact, dead-code, diagram, refactor-plan, onboarding-tour, graph-status, index, update)
- Three sub-agents (dependency-analyzer, find-usages, explain-service) with explicit tool/model choices
- Full hook suite: SessionStart, UserPromptSubmit (staleness tracking), PreToolUse (grep/glob/bash augmentation), PostToolUse (impact analysis), Stop, PreCompact (compaction recovery), SubagentStop (transcript audit), Notification
- Git PAT auth via `/auth` (delegates to CLI; token never transits the chat), configurable index timeout, statusline
- Plugin test suite (89 tests: hooks, skills, common, statusline)

**Agent (opentraceai)**
- New MCP tools: source_read, source_grep, find_usages, impact_analysis, repo_index, fts_search
- `opentraceai diff` for impact analysis on uncommitted changes
- Rank Function/Class above File in find_usages and keyword_search; src-layout import resolution
- Version bump 0.5.0 → 0.6.0

**Tooling**
- `scripts/check_versions.py` version consistency check


ghost (commented):
## 🔧 Changes Requested

This PR is a solid expansion of the OpenTrace agent's capabilities. The architectural shift toward modular hooks and skill-based routing in the Claude Code plugin is a great move for maintainability and complex agentic workflows.

I've flagged a few issues that need to be addressed before merging, primarily a security vulnerability regarding temporary file handling and some fragile parsing logic in the graph store. See my inline comments for the specific details.

### Broader observations

*   **Performance of `impact_analysis`**: The current implementation in `mcp_server.py` performs a separate BFS traversal for every symbol defined in a file. For large files or deep dependency trees, this could become a bottleneck for the model's response time. In a future iteration, you might consider a single Cypher query that finds the union of dependents for all symbol IDs in the file to reduce round-trips.
*   **CLI Tooling**: The addition of `source-search`, `diff`, and the enhanced `mcp` server provides a much richer set of primitives for both humans and agents. The use of `uvx` as a fallback in the plugin hooks is a robust choice for handling various installation environments.

<details>

1. **Grep/Read pattern extraction**: `plugins/claude-code/scripts/_common.py:273`
   The extraction logic is quite comprehensive, but it's worth noting it might miss more complex shell redirections or process substitutions. Given this is for augmentation context, the current "best-effort" approach is likely sufficient.

2. **FTS Over-fetching**: `agent/src/opentrace_agent/store/graph_store.py:157`
   The multiplier approach for repo-scoped FTS is a clever workaround for Kuzu's global ranking. Just keep an eye on memory usage if `limit` is set very high by a user.

</details>

plugins/claude-code/scripts/_common.py
```diff
@@ -0,0 +1,595 @@
...
+# Per-UID directory so multi-user systems don't collide.
+# ---------------------------------------------------------------------------
+
+_UID = getattr(os, "getuid", lambda: "shared")()
+CACHE_DIR = Path(os.environ.get("TMPDIR", "/tmp")) / f"opentrace-claude-hooks-{_UID}"
+BRIEFING_CACHE_PATH = CACHE_DIR / "briefing.json"
+CONTEXT_CACHE_PATH = CACHE_DIR / "context.json"
+# Per-workspace edited-file tracker; keys are sha256 of (workspace_root|path)
+# so multi-repo sessions don't collide. Values are {path, mtime, ts}.
+STALENESS_CACHE_PATH = CACHE_DIR / "staleness.json"
```
ghost, line 52: 🚨 This is a security vulnerability (Insecure Temporary File).

`CACHE_DIR` uses a predictable name in a world-writable directory (`/tmp`) and the plugin hooks write to files within it (e.g., `BRIEFING_CACHE_PATH.write_text`) without ensuring they aren't malicious symlinks. 

A local attacker could pre-create these files as symlinks to a sensitive file owned by the user (like `~/.ssh/authorized_keys`), and the next time the plugin runs, it would overwrite that file with JSON content.

Consider using a more secure location like `~/.cache/opentrace/` or using `tempfile` correctly (though persistence across processes might be needed). At a minimum, use `os.makedirs(..., mode=0o700, exist_ok=True)` and check ownership if the directory exists.

agent/src/opentrace_agent/cli/mcp_server.py
```diff
@@ -182,4 +534,587 @@ def get_stats() -> str:
...
+        startLine: int = 0,
+        endLine: int = 0,
+    ) -> str:
+        """Read source code from any indexed repository.
+
+        Provide either ``nodeId`` (preferred — looked up via the graph to
+        recover its file path and line range) or a ``path`` relative to a
+        repo root (with optional ``repo`` to disambiguate).  ``startLine`` /
+        ``endLine`` clip the returned slice; leave 0 to read the whole file.
+        Returns ``cat -n``-style numbered output.
```
ghost: ⚠️ The docstring says it returns numbered output, but it doesn't when both `startLine` and `endLine` are 0 (line 779 calls `_read_file_slice` which returns raw content if both are None). It might be better to always return numbered output for consistency if that's the intended behavior.

plugins/claude-code/scripts/_common.py
jatinot: Addressed in c0923dd. All four cache writes now go through a `_ensure_cache_dir()` guard (the only `mkdir` left is inside it):

- creates the dir `mode=0o700` so no other user can drop files/symlinks inside it;
- on an existing dir, `lstat` + `st_uid` reject a symlinked or foreign-owned directory by raising `OSError` — callers already wrap writes in `try/except OSError`, so a hijacked dir degrades to "no cache" instead of clobbering a target;
- `os.chmod(..., 0o700)` re-tightens perms if it pre-existed.

This closes the symlink-clobber path: the attack needs a hostile symlink *inside* the cache dir, which a 0700 dir we own cannot contain. Kept the path under `TMPDIR` (rather than `~/.cache`) so the statusline script and tests still resolve the same location; the guard is the mitigation you noted as acceptable.

PR #74, opened 2026-03-17
feat(ui): responsive toolbar redesign with three-tier breakpoints

Umaaz:
## Add mobile-responsive hamburger menu
🆕 **New Feature** · ✨ **Improvement**

Adds a hamburger menu to the header toolbar that appears on mobile viewports (≤768px). The menu toggles visibility of header controls (search, theme selector, buttons) in a vertical stack below the logo.

### Complexity
🟢 Low · `2 files changed, 88 insertions(+)`

Adds a self-contained mobile UI pattern (burger menu + collapsible section) to a single component. The change touches only presentation logic with straightforward show/hide behavior and CSS media queries.

### Tests
🧪 No tests included — adds UI behavior without corresponding test coverage.

### Review focus
Pay particular attention to the following areas:

- **Media query breakpoint** — verify 768px is the right threshold for this application's mobile layout
- **Accessibility** — check that aria-label and aria-expanded provide adequate screen reader context

Issue #67, opened 2026-03-17
Add configuration quickstart with required tools


Umaaz:
The setup documentation is missing a short quickstart for required tooling, which made it unclear what needs to be installed and where.

Include a brief configuration section that notes which tools are needed by subdirectory:

* `node` / `npm` for `ui/`
* `uv` for `agent/`
* Go tools for `api/`
* `protoc` and gRPC tools for `proto/`

It should also cover basic version checks and common install commands for supported platforms.

PR #399, opened 2026-06-18
feat(ui): move store + embedder off main thread, refresh graph defaults

dapywell:
## Web Worker offloading and server-side background indexing
🆕 **New Feature** · ✨ **Improvement** · 🐛 **Bug Fix** · ⚡ **Performance**

Moves the graph store and ML embedder to Web Workers to ensure a 60fps UI during indexing and adds server-side background indexing for remote repositories. It also introduces significant performance optimizations via bulk CSV imports and stratified sampling for large graphs.

- **Non-blocking UI**: Store and Embedder operations are now asynchronous and run off the main thread.
- **Server Indexing**: The agent can now clone and index remote URLs in the background via a new REST API.
- **Bulk Import**: Replaces per-row inserts with `COPY FROM` (CSV) for both the Python agent and browser WASM store.
- **Rendering**: Smooths out 3D auto-rotation and decouples community visual settings from global labels.

### Complexity
🟠 High · `49 files changed, 4249 insertions(+), 417 deletions(-)`

This PR introduces a foundational shift in the project's architecture by moving the core data engine and inference pipeline into Web Workers. It significantly expands the system boundary by adding background job management and remote repository indexing to the agent. These changes require careful review of cross-thread communication (main thread to worker), thread-safe database swapping in the agent, and a new bulk-ingestion protocol that replaces standard Cypher statements.

### Tests
🧪 Existing unit tests for the store and indexer have been updated to handle the new asynchronous interfaces.

### Review focus

- **Worker communication** — verify the `WorkerGraphStore` proxy correctly handles promise resolution, sequence matching, and error propagation from the worker.
- **Database Swap Safety** — check the logic in `serve.py` that closes the store, swaps the DB file, and reopens it to ensure readers don't hit locked or missing files.
- **Bulk Import Integrity** — review the `COPY FROM` pre-flight deduplication and FK filtering to ensure it accurately mirrors the previous `MERGE` behavior.

---

## Folded in after the initial review

Two further bodies of work were merged into this branch after the first review pass.

### Chat: API-key entry auto-detects the provider
- Replaced the up-front provider picker with a single key field that infers the provider from the key prefix (`sk-ant-…` → Anthropic, `AIza…` → Gemini, `sk-…` / `sk-proj-…` → OpenAI). Unrecognized keys fall back to a manual picker; local-model mode is unchanged.
- The "how to get a key" guide now lists all supported providers alphabetically behind a **"Show me how to get a key for:"** selector, so it no longer reads as single-provider.

### Merged `fix/minor-bug-sweep` — 17 standalone bug fixes
Independent fixes, each with its own tests; none overlapped the worker/indexing work (the one merge conflict, in `useGraphData.ts`, combined both sides' latest-wins guards).

**Agent (Python)**
- Parse multi-line PEP 621 `[project]` dependencies, including `extras` brackets
- Correct TypeScript `implements` + class-field extraction
- Only attach GoDoc comments that are directly adjacent (skip blank-line-separated)
- Match the full Go package directory, not just the last path segment
- Clean up the temp clone dir when `git clone` fails
- `serve` opens the DB read-only to avoid blocking a concurrent index
- Honor `bench --verbose`; stop `augment` from swallowing errors

**UI**
- Keep dots in GitHub repo names in `parseGitHubUrl` (`react.dev` is no longer truncated to `react`)
- Key TypeScript imports by local binding name, not file basename
- Controlled repulsion slider; bounded LLM-summary concurrency
- Unstick the AddRepoModal submit button after a failed index
- Discard stale overlapping `loadGraph` results (latest-wins guard)
- `runWithConcurrency` survives per-item failures; `truncate()` always emits valid JSON; `VectorIndex.addVector` no longer mutates the caller's array

All checks green on the combined branch — UI: `build` ✓, `eslint` 0 errors, `vitest` **932 passed**; agent: `ruff` clean, `pytest` **808 passed**. Browser-observable UI fixes were also verified live in the preview.

---

<details>

### Worker Architecture (Plans D & E)
The core `LadybugGraphStore` and `Transformers.js` embedder have been moved into dedicated Web Workers. 
- **WorkerGraphStore**: A main-thread proxy that communicates over `postMessage`. It uses a sequence-number protocol to match requests to promises.
- **Embedder Shim**: Since the store worker needs an embedder to build its vector index, it uses a shim that bounces embedding requests back to the main thread (which owns the `WorkerEmbedder`), avoiding the overhead of loading the ONNX model in multiple workers.

### Bulk Import Optimization
To handle larger codebases, ingestion now uses a "fast-path" `COPY FROM` strategy:
- **Agent (Python)**: Streams data through temporary CSV files to leverage LadybugDB's native bulk insert.
- **Browser (WASM)**: Generates typed CSV strings and uses the virtual filesystem to execute `COPY` commands.
- **Safety**: Both implementations include pre-flight deduplication and a fallback to per-row `MERGE` statements if the bulk operation fails.

### Server-Side Indexing
The `opentrace_agent` now supports a full background indexing lifecycle:
1. **POST /api/index-url**: Triggers a background thread to clone/update a repo and run the indexing pipeline.
2. **Job Tracking**: The UI polls `/api/index-progress/{id}` for live status (nodes/edges created, current file).
3. **Atomic Swap**: Once indexing completes in a `.staging` file, the agent closes the live store, renames the file, and reopens the connection.

</details>


ui/src/store/ladybugStore.ts
```diff
@@ -935,10 +937,107 @@ export class LadybugGraphStore implements GraphStore {
...
+          nodeIdSet.add(id);
+        }
+      }
+
+      // Fetch all edges, keep those with both endpoints in the sample,
+      // stop at the edge cap.
+      const relRows = await this.query(
+        `MATCH (a)-[r:RELATES]->(b) ` +
+          `RETURN a.id AS source, b.id AS target, r.type AS type, r.properties AS properties`,
+      );
```
ghost: 🚨 This will OOM the worker on large repositories. Fetching every edge in the database (`MATCH (a)-[r:RELATES]->(b)`) into memory just to filter them in JS is dangerous when `totalEdges` can be 100k+. 

Since you have the `nodeIdSet` from the stratified sampling above, it's safer to query for edges specifically between those nodes, or at least apply a `LIMIT` to the edge query as well (similar to Strategy B).

agent/src/opentrace_agent/cli/serve.py
```diff
@@ -71,18 +117,44 @@ async def fetch_graph(request: Request) -> JSONResponse:
...
-            stats = store.get_stats()
+            stats = cur.get_stats()
             for ntype in stats.get("nodes_by_type", {}):
-                all_nodes.extend(store.list_nodes(node_type=ntype, limit=limit))
-                if len(all_nodes) >= limit:
+                all_nodes.extend(cur.list_nodes(node_type=ntype, limit=max_nodes))
+                if len(all_nodes) >= max_nodes:
                     break
-            all_nodes = all_nodes[:limit]
+            all_nodes = all_nodes[:max_nodes]
```
ghost: ⚠️ This will return a biased sample of nodes (mostly from the first few types iterated) instead of a stratified sample like the browser store does. If the first node type has more than `max_nodes` entries, the initial graph view will only show that one type.

ui/src/store/createLadybugStore.ts
```diff
@@ -14,17 +14,19 @@
...
+/** Returns a singleton store backed by the in-browser WASM engine.
+ *  The LadybugGraphStore class is hosted inside a Web Worker; this returns
+ *  a thin main-thread proxy implementing the same GraphStore interface. */
+export function createLadybugStore(): WorkerGraphStore {
   if (!singletonStore) {
-    singletonStore = new LadybugGraphStore();
+    singletonStore = new WorkerGraphStore();
   }
   return singletonStore;
 }
```
ghost, line 32: ⚠️ Should this worker be disposed when switching from 'local' to 'server' mode? Currently, once initialized, it stays alive in memory for the lifetime of the page.

ui/src/store/ladybugStore.ts
```diff
@@ -935,10 +937,107 @@ export class LadybugGraphStore implements GraphStore {
       );
     }
 
-    // For large graphs, fetch edges first then their endpoint nodes.
-    // This guarantees every edge has both endpoints in the result.
+    // Two large-graph strategies, depending on which cap is binding:
+    //
+    //  (a) Nodes don't fit (totalNodes > maxVisNodes): fetch edges first
+    //      then their endpoint nodes. Picks the most-connected subgraph
```
ghost: ⚠️ This comment contradicts the implementation below. The code actually performs stratified sampling of nodes first, then fetches edges.

```suggestion
    //  (a) Nodes don't fit (totalNodes > maxVisNodes): fetch a stratified
    //      sample of nodes first, then keep only edges that connect them.
```

ghost (commented):
## 🔧 Changes Requested

Overall, this is a solid set of changes that significantly improves the responsiveness of the application during indexing. Moving the store and embedder to workers, combined with the `COPY FROM` bulk import path, addresses several major performance bottlenecks.

I've left a few inline comments, but the most important one is a potential OOM risk in Strategy A of the graph sampling logic. Fetching every relationship in a large database into a JS array to filter it will likely crash the worker on bigger repositories.

See inline comments for details on that and a few other points regarding the agent's sampling logic and worker lifecycle. Happy to approve once the edge sampling is hardened.

<details>

1. **Chunked `storeSource`**: `ui/src/store/workerStore.ts:169`
   Chunking the file list to avoid blocking the main thread during structured cloning is a great find and a smart fix for large repos.

2. **DuckDB buffer size**: `ui/src/store/ladybugStore.ts:559`
   The comment regarding `bufferPoolSize` and WASM linear memory is very helpful context for anyone debugging memory issues in the future.

3. **3D Auto-rotate**: `ui/src/appComponents/GraphViewer.tsx:700`
   Good attention to detail in the reset logic, ensuring the imperative renderer state is fully synchronized even when React props don't technically change.

</details>

dapywell:
## Detailed fix list

The PR description covers the high-level themes; here's the per-fix breakdown for anyone walking through the diff. Numbers follow the `Fix #N` markers already in the source comments through #56, with #57+ assigned sequentially to the work added during this session.

### Renderer & display

- **Fix #7** — Edges-visibility toggle that hides the edge layer at the renderer level (preserving positions) instead of removing edges from data.
- **Fix #8** — React to layout-mode flips in 3D mode without leaving the projection collapsed at the previous mode's radius.
- **Fix #9** — Deferred `set3DMode` so re-entering 3D after a layout change re-projects against fresh positions.
- **Fix #22** — Push persisted zoom/label-scale settings into the renderer on mount instead of waiting for a slider touch.
- **Fix #27** — World-space label gap so labels sit just outside the sprite edge at deep zoom (previously slid inside the circle).
- **Fix #32** — Community-label alpha fade by zoom (baseline).
- **Fix #33** — Community labels track sprite (projected) positions in 3D so they ride rotation instead of staying at un-projected base coords.
- **Fix #34** — Fold community-centroid accumulation into the projection loop so wayfinders re-aim every frame at zero extra cost.
- **Fix #35** — Sticky-first ordering when the cull re-runs (keeps prior decision when geometry doesn't force a change).
- **Fix #36** — Freeze community label visibility after layout settles, to stop labels blinking during 3D rotation.
- **Fix #38** — No label cull on rotate/release; the visible set is decided by zoom and stays stable across rotation/pan/release.
- **Fix #42** — Symmetric refit when toggling community clusters (replaced an earlier direction-aware refit that made compact look like a tiny ball).
- **Fix #43** — Piecewise counter-scale formula: constant screen size until threshold, then shrinks past it.
- **Fix #47** — Tightened shrink threshold so wayfinders drop in size early as the user zooms past fit-to-screen.
- **Fix #48** — Freeze cull only after `layoutSettled` so layout-add-time origin-clustered centroids don't lock in a bad visible set.
- **Fix #49** — Viewport-margin pre-filter and `MIN_SPRITE_SCREEN_RADIUS` gate so dust-sized nodes don't claim label slots.
- **Fix #52** — Decouple community label visibility from `layoutMode`. Toggling no longer reflows the graph; the compact↔spread toggle lives separately.
- **Fix #54** — Held auto-rotation until `layoutSettled` (later reverted by Fix #60 once Plan D + E removed the stutter source it was working around).

### Workers / threading

- **Fix #55 — Plan E** — Moved the `@huggingface/transformers` embedder into a Web Worker. Main-thread `WorkerEmbedder` proxy implements the `Embedder` interface unchanged.
- **Fix #56 — Plan D** — Moved the entire `LadybugGraphStore` into a Web Worker. `WorkerGraphStore` proxy keeps the `GraphStore` interface unchanged; cross-worker embed requests bounce through main thread to avoid duplicating the ONNX model.
- **Fix #57** — Moved `localStorage` reads out of `LadybugGraphStore.initModule()` since Web Workers can't access `localStorage`. Vis caps are now read on main and pushed in via `setLimits()` after `ensureReady()`.
- **Fix #58** — Chunked `storeSource()` across `postMessage` (200 files per chunk). A single ~hundreds-of-MB structured-clone was freezing main for seconds on large repos.

### Settings & UX

- **Fix #59** — Renamed the mislabeled PhysicsPanel toggle "Community clusters" → "Community labels" (it only controls labels; the real cluster-force toggle is in PixiControlPanel).
- **Fix #60** — Removed the `&& layoutSettled` gate on 3D auto-rotation (reverses #54). With Plan D + E, the original stutter is gone, and the gate was suppressing rotation entirely on graphs where d3-force never reports settled.
- **Fix #61** — Apply counter-scale immediately when community labels are toggled back on, so they appear at correct screen size instead of stale-scaled until the next pan/zoom.
- **Fix #62** — Apply counter-scale immediately after `rebuildCommunityLabels()` creates new `Text` objects (same root cause as #61 but in the post-Louvain create path).
- **Fix #63** — Lowered `MIN_SPRITE_SCREEN_RADIUS` from 4 → 2.5 so anchor nodes get labels at lower zoom levels.
- **Fix #69** — Centralized panel defaults in `GRAPH_SETTING_DEFAULTS` as a single source of truth.
- **Fix #70** — New `resetSettings()` that resets every panel setting back to its default and clears `localStorage`.
- **Fix #71** — "Reset graph" button extended: now calls `resetSettings()` **and** pushes every default into the renderer (`setLayoutMode`, `setZoomSizeExponent`, `set3DMode`, etc.) so the canvas reflects them immediately. Previously it only reheated physics + reset camera.
- **Fix #72** — Default `layoutMode` changed from `'spread'` → `'compact'`.
- **Fix #73** — Default `pixiZoomExponent` changed from `0.8` → `0.25` (25% zoom scaling).
- **Fix #74** — Default `zoomOnSelect` changed from `true` → `false`. Clicking a node updates the side panel without moving the camera.

### Sampling / graph fetch

- **Fix #64** — When `total_nodes > maxVisNodes`, replaced the arbitrary-slice strategy (~0.5% edge survival) with **stratified per-type sampling** + overflow redistribution. Functions/Classes/Files are now all represented instead of the sample being dominated by whichever type comes first in storage order.
- **Fix #65** — Nodes-first strategy when only the edge cap is exceeded — caller's full node cap is honored.
- **Fix #66** — Master "Communities" toggle. When off, Louvain doesn't run at all and the three sub-controls (colors, labels, pull) are visually disabled.
- **Fix #67** — Pre-flight OOM detection in `BrowserJobService` so indexing fails fast with a "switch to server mode" message instead of crashing mid-flush.
- **Fix #68** — Tried setting `bufferPoolSize` to 512 MB on `lbug.Database`. Broke ingest of even small repos because WASM heap is shared between buffer pool and CSV staging. **Reverted** with an explanatory comment.

### Agent (Python)

- **Fix #75** — New REST endpoints (`POST /api/index-url`, `GET /api/index-progress/{job_id}`, `POST /api/clear`) for browser-driven indexing.
- **Fix #76** — `_run_indexing_pipeline` accepts an `on_event` callback so REST/MCP can stream live progress.
- **Fix #77** — Extractor parity tweaks (Go / Python / TypeScript).

---

## Recommended follow-ups (flagged by security review)

Not blocking this PR but should land soon, ideally in a follow-up:

- **Fix #78 — CSRF hardening**. Replace `allow_origins=["*"]` in [serve.py:651-658](../blob/ui-fixes/agent/src/opentrace_agent/cli/serve.py#L651-L658) with `allow_origin_regex=r"https?://(localhost|127\.0\.0\.1)(:\d+)?"` and tighten `allow_methods` / `allow_headers`. The wide-open CORS is pre-existing but the new `/api/clear` and `/api/index-url` endpoints make it destructively exploitable from any tab.
- **Fix #79 — Path traversal hardening**. In `_run_index_url_job`, validate `repoUrl` with `urlparse` (require `http`/`https` scheme, non-empty host, no `..` in path), and verify `clone_root.resolve().is_relative_to(Path.home() / ".opentrace" / "repos")` before any `shutil.rmtree`. Crafted `repoUrl` like `https://attacker.example/x/..` can otherwise wipe the entire cache directory.

---

## Verification summary

- `npx tsc --noEmit` clean across the repo.
- `npx vitest run` — 139 / 139 tests pass (store, graph hooks, pipeline embed stage).
- Manual: indexed express, podinfo, and tempo in browser mode; exercised search (forces the cross-worker embed-request bounce); toggled the master Communities switch; clicked "Reset graph" and confirmed every panel setting reverts + canvas reflows. 3D rotation runs continuously instead of pausing in bursts during the embed phase.

jatinot:
## Review comments + CI — addressed

**CI fix:** the failing `Agent (Python)` check was `ruff format --check` (3 files: `main.py`, `serve.py`, `graph_store.py`). Ran `ruff format`; `ruff check`, `vitest` (932), and `pytest` (809) are all green locally now.

Most inline comments were posted against intermediate commits and are **already resolved** in the merged code. Verified each against `HEAD`:

| Comment | Status |
|---|---|
| `serve.py` import sorting (I001) | ✅ already sorted |
| `serve.py` index-vs-index TOCTOU | ✅ atomic check-and-insert under `index_jobs_lock` |
| `serve.py:155` "biased sample" | ✅ already stratified (ascending-count `even_share` allocation) |
| `ladybugStore.ts` edge-fetch OOM | ✅ endpoint filter + `LIMIT maxVisEdges` pushed into the Cypher query |
| `useGraphData.ts` opposite error channel | ✅ both branches clear the other channel |
| `GraphInteractionProvider` `communitiesEnabled` persistence | ✅ read-merge-write into `graph-settings` |
| `import_analyzer.py` Go module boundary | ✅ strips module prefix + full-dir match |
| `PixiRenderer.ts` unused `_duration` | ✅ intentional underscore param (UI CI passes) |

**Genuinely open — fixed in `c4e17bc`:**
- `manifest_parser.py`: multi-line PEP 621 dependency arrays now capture **single-quoted** entries too, not just double-quoted (+ test).
- `chat/providers.ts`: tightened Gemini detection to exactly `AIza` + 35 chars to avoid false-positive auto-selection (+ tests).
- `AddRepoModal.tsx`: added `pat`/`ref` to the retry-reset effect deps so editing the token/branch also clears a stuck submit.

The two security follow-ups (#78 CORS, #79 path-traversal) remain intentionally out of scope for this PR, as noted — happy to do them in a fast follow-up.

ui/src/components/pixi/PixiRenderer.ts
```diff
@@ -1659,16 +2252,81 @@ export class PixiRenderer {
    * the timer on every call (classic debounce) would starve the fit
    * indefinitely because sim ticks arrive faster than the throttle window.
    */
-  scheduleAutoFit(duration = 200): void {
+  scheduleAutoFit(_duration = 200): void {
```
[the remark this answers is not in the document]
ghost: This was addressed in c4e17bc by prefixing the unused parameter with an underscore and updating the ESLint configuration to allow this pattern for API compatibility. Resolving this one.

ghost:
## ✅ Review Update — Approved
> 22 new commits: 7ad4159, a561cf4, ae21473, 8efb928, 83de9d7, f604d8d, 324fd21, 51ebad8, 52d5e1e, 2b7ac0d, eb86805, a49e898, 2ec9187, 143202b, 90891ca, 7c55ab2, 4ac2d80, 0108b81, 75bbd45, c82e3e8, c4e17bc, 073aaf8

All the feedback from the initial review has been addressed.

The move to stratified sampling ensures a much more balanced initial graph view, and pushing the edge filters into the Cypher queries solves the OOM risk on large repos cleanly. I also appreciate the concurrency guards in `serve.py` and the serialization of embed requests in the worker — those make the system significantly more robust.

I've resolved the remaining discussion threads. I saw your note about the CORS and path traversal hardening — those are definitely worth tackling in a fast follow-up as planned.

This is good to go.

dapywell:
@jatinot — requesting your review. All earlier inline threads are resolved and CI is green; the PR just needs a formal approval to clear `REVIEW_REQUIRED` before merge. Thanks!

jatinot (approved):

PR #401, opened 2026-06-25
fix(ui): run LadybugDB engine on main thread to avoid nested worker timeout

jatinot:
## Fix Safari/WebView timeout by moving LadybugDB to main thread
🐛 **Bug Fix** · ✨ **Improvement** · ♻️ **Refactor**

Moves the LadybugDB engine from `storeWorker` to the main thread to prevent nested worker initialization failures. 

This fix resolves 60-second timeouts in Safari and WebViews where worker-in-worker support is unreliable. Heavy data processing remains in the worker, while raw engine operations are proxied via a new RPC interface.

### Complexity
🟡 Moderate · `6 files changed, 413 insertions(+), 157 deletions(-)`

This PR restructures the communication between the application's main thread and its background store worker. While it follows existing architectural patterns for shimming services into workers, it touches the core database lifecycle and changes infrastructure-level security headers in the Vite configuration.

### Tests
🧪 Existing unit tests were refactored to use a mock engine, maintaining coverage of the store's query queueing and state management.

### Review focus

- **RPC Bridge Integrity** — verify that query results and errors are correctly serialized and returned across the worker boundary.
- **Zero-copy Transfers** — ensure CSV buffers in `fsWrite` are correctly included in the `postMessage` transfer list to avoid performance regressions.
- **COEP Header Alignment** — check that the shift to `require-corp` in the dev server doesn't block local assets in environments that were relying on `credentialless`.

### DIFF
--- a/ui/src/store/__tests__/ladybugStore.test.ts
+++ b/ui/src/store/__tests__/ladybugStore.test.ts
@@ -30,24 +30,29 @@ describe('REL_PAIRS', () => {
   });
 });
 
-// Helper: build a store with WASM init bypassed and a mock conn that
-// resolves every query() call into a trivial result.
-function makeStoreWithMockConn() {
-  const store = new LadybugGraphStore();
+// Helper: build a store with WASM init bypassed and a mock engine whose
+// query()/exec() both funnel through one spy, so assertions can inspect every
+// Cypher statement the store dispatched (regardless of rows-vs-no-rows).
+function makeStoreWithMockEngine() {
+  const engineCall = vi.fn().mockResolvedValue([]);
+  const engine = {
+    init: vi.fn().mockResolvedValue(undefined),
+    query: (cypher: string) => engineCall(cypher),
+    exec: (cypher: string) => engineCall(cypher),
+    fsWrite: vi.fn().mockResolvedValue(undefined),
+    fsUnlink: vi.fn().mockResolvedValue(undefined),
+    close: vi.fn().mockResolvedValue(undefined),
+  };
+  const store = new LadybugGraphStore(engine);
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const s = store as any;
   s.ready = Promise.resolve();
-  const connQuery = vi.fn().mockImplementation(async () => ({
-    getAllObjects: async () => [],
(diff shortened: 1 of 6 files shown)


ghost (commented):
## ✅ Approved

The architectural shift to move the LadybugDB engine to the main thread is a solid fix for the nested worker timeout issue. The RPC implementation between the store worker and the main thread is clean and correctly uses zero-copy transfers for bulk data.

I've noted one important race condition in the `flush()` logic regarding shared filenames in the virtual filesystem. While the store handles many sequential guarantees, the concurrent use of fixed paths like `/nodes_${type}.csv` could lead to data corruption or ingest failures if multiple batches are flushed in close succession.

Aside from that, the changes look great and significantly improve browser compatibility.

<details>

1. **Memory reporting**: `ladybugStore.ts:831` — Returning `-1` for WASM memory is a reasonable trade-off given the architectural change, but it might be worth adding a comment to the UI where this is displayed so users aren't confused by "Unknown" memory usage.
2. **Dispose logic**: `workerStore.ts:274` — The double-close approach (RPC followed by worker termination and a direct engine close) is excellent "belt-and-suspenders" engineering.
</details>

ghost:
## ✅ Review Update — Approved
> 1 new commit: d8ebc4e

Nice work on the follow-up fix in d8ebc4ef — that's a great catch. Using unique temp filenames for the `COPY FROM` operations makes the ingest path much more robust against VFS collisions, especially since those raw engine ops are now crossing the worker boundary.

The overall approach of moving the LadybugDB engine to the main thread is a solid fix for the Safari/WebView timeout issues. I've verified that the RPC bridge correctly uses transfer lists for the CSV buffers to maintain zero-copy performance.

The alignment of the COEP header to `require-corp` in `vite.config.ts` also looks good; keeping dev and prod consistent is definitely the right move for Safari support.

One minor thing to keep in mind: since the main-thread engine persists even if the store worker restarts, a `copySeq` starting at 0 might still hit orphaned files from a previous session. A random suffix or UUID would be bulletproof there, but this is already a significant improvement over the fixed paths.

This is ready to go.

dapywell (approved):

PR #402, opened 2026-06-26
feat(ui): add graph.css export for graph components only

callumyuille:
## Export graph-only CSS for custom design systems
🆕 **New Feature** · ✨ **Improvement** · 📝 **Docs**

This PR adds a `graph.css` export containing only core graph component styles. 

It allows consumers to use OpenTrace components with their own design tokens while avoiding the global resets and themes included in the main stylesheet.

### Complexity
🟢 Low · `3 files changed, 141 insertions(+), 2 deletions(-)`

Scope is limited to a new build-time script and package metadata. No functional code or existing styles are modified, and the new output is isolated via CSS layers.


ui/scripts/build-graph-css.mjs
```diff
@@ -0,0 +1,127 @@
...
+ */
+`;
+
+// Strip the leading Apache license block comment so the generated file carries
+// a single header rather than one per source file.
+function stripLicenseHeader(css) {
+  return css.replace(
+    /^\s*\/\*[\s\S]*?Copyright 2026 OpenTrace Contributors[\s\S]*?\*\/\s*/,
+    '',
+  );
```
ghost, line 90: ⚠️ The copyright year in the stripping regex is hardcoded to 2026. Making it a wildcard ensures this script won't break or stop stripping headers when the year is updated in the source files.

```suggestion
function stripLicenseHeader(css) {
  return css.replace(
    /^\s*\/\*[\s\S]*?Copyright \d{4} OpenTrace Contributors[\s\S]*?\*\/\s*/,
    '',
  );
}
```

ghost (commented):
## ✅ Approved

This is a solid addition that makes the library much friendlier for consumers with their own design systems. Wrapping the panel styles in the `@layer opentrace` cascade layer is a great defensive choice to prevent the library's styles from overriding host application styles unexpectedly.

The build script is straightforward and integrates well into the existing `build:lib` flow. See my inline suggestion regarding the license-stripping regex for a minor improvement in resilience.

One detail to keep an eye on: the script currently only bundles CSS from `src/components/panels/`. If the `GraphCanvas` or other core components start requiring external CSS files (rather than inline styles or canvas-only logic), those will need to be added to the glob in the build script. Based on the current structure, this covers the primary UI components that consumers are likely to style.

ghost:
## 🔄 Review Update
> 1 new commit: 6363db9

The new commit updates the copyright stripping regex as suggested, which is great. However, it looks like an extra closing brace was accidentally introduced on line 92 of `ui/scripts/build-graph-css.mjs`, which will break the build.

Once that syntax error is cleaned up, this is ready to land.

ghost:
## ✅ Review Update — Approved
> 2 new commits: 6363db9, e952e75

The syntax error in the build script is gone, and the copyright stripping regex looks much more robust now with the wildcard year. Everything looks solid—this is ready to go.

AnysSido (approved):

PR #403, opened 2026-06-28
feat(ui): Three.js graph renderer — 3D layouts, view presets, ambient motion

jatinot:
## Upgrade to Three.js renderer and off-thread graph management
🆕 **New Feature** · ✨ **Improvement** · ♻️ **Refactor** · ⚡ **Performance** · 🧪 **Tests**

Replaces the PixiJS renderer with a 3D-capable Three.js engine and moves the graph store to a background worker. 

Introduces view presets (Planet, Nebula, Tree), progressive graph loading for massive datasets, and real-time chat traversal animations.

### Complexity
🔴 Complex · `54 files changed, 8929 insertions(+), 7183 deletions(-)`

This is a foundational architectural shift that replaces the project's primary visualization layer. It introduces complex multi-worker orchestration (Store Worker, Layout Worker, and Main Thread) and implements a high-performance rendering pipeline using custom GLSL shaders. The changes span from low-level database interactions to high-level React component orchestration, requiring a deep understanding of the entire system's data flow.

### Tests
🧪 Includes updated unit tests for the LadybugStore and coverage for the new progressive loading logic.

### Review focus

- **Worker Coordination** — Check for race conditions in `storeWorker.ts` and `ThreeGraphCanvas.tsx` when switching repos or clearing state while workers are busy.
- **Shader Correctness** — Verify the state-masking logic in `nodeMaterial.ts` correctly handles combined highlight/dim/visible states.
- **Traversal Backlog** — Ensure the `TRAVERSAL_MAX_BACKLOG_MS` logic in `ThreeRenderer.ts` prevents animation drift during rapid agent tool calls.
- **Resource Disposal** — Confirm that the WebGL contexts and Worker instances are properly terminated when the graph component unmounts.

---

<details>

### Architecture & Approach
The migration replaces the imperative Pixi drawing loop with a declarative-ready **Three.js** system designed for 100k+ node performance. 

*   **Three.js Renderer**: Uses `THREE.Points` for nodes and `THREE.LineSegments` for edges. Node highlighting, dimming, and sizing are handled in custom shaders (`nodeMaterial.ts`) to avoid per-frame CPU loops over the scene graph.
*   **Store Worker**: The `LadybugGraphStore` and the embedder shim now live in `storeWorker.ts`. This moves all Parquet export/import, CSV generation, and BM25 index building off the main thread, keeping the UI at 60fps during heavy background indexing.
*   **3D Force Layout**: The layout engine has been upgraded to `d3-force-3d` within a dedicated worker. It supports seamless switching between 2D and 3D dimensionality by rebuilding the simulation in-place while preserving existing node coordinates.

### Traversal Logic
The "Chat Traversal" feature in `ThreeRenderer.ts` uses an internal animation timeline. As the agent returns tool results, the renderer computes the shortest path from the "discovered" graph to the new results and glides a pulse along the edges. If tool results arrive faster than the animation can play, it includes a fast-forward mechanism to prevent the visual state from lagging behind the chat history.

</details>


ui/src/components/three/ThreeRenderer.ts
```diff
@@ -0,0 +1,3905 @@
...
+    ).needsUpdate = true;
+
+    // Edges connect the live (flying) node positions; alpha ramps in with the
+    // later-born endpoint so an edge fades in as its child node lands. Same
+    // filters as updateEdgeAlpha so the build matches the user's current view.
+    if (this.edgeGeometry) {
+      const ep = this.edgePosArray;
+      const ea = this.edgeAlphaArray;
+      const idx = this.nodeIdToIndex;
+      for (let k = 0; k < this.edges.length; k++) {
```
ghost, line 5024: ⚠️ This iterates over all nodes and edges EVERY frame (60fps) on the main thread. For a graph with 100k nodes and 300k edges, this will almost certainly hang the browser tab during the animation. 

While it's an opt-in/special animation, we should ideally move this logic to a vertex shader or at least avoid per-edge JS work here. At minimum, we should pre-calculate the edge endpoints to avoid the map lookups at line 3482-3483.

ui/src/components/workers/forceLayout3dWorker.ts
```diff
@@ -0,0 +1,1628 @@
...
+    const cz = new Map<number, number>();
+    const count = new Map<number, number>();
+    for (const node of nodes) {
+      const c = comms[node.id];
+      if (c === undefined) continue;
+      cx.set(c, (cx.get(c) ?? 0) + (node.x ?? 0));
+      cy.set(c, (cy.get(c) ?? 0) + (node.y ?? 0));
+      if (is3D) cz.set(c, (cz.get(c) ?? 0) + (node.z ?? 0));
+      count.set(c, (count.get(c) ?? 0) + 1);
+    }
```
ghost, line 382: ⚠️ This force function is called every tick of the simulation. For a graph with 100k nodes, we are allocating 4 new `Map` objects and performing 400k map operations every tick. This will create massive GC pressure and significantly slow down the layout.

We should pre-allocate typed arrays for the centroids or use a more efficient way to store community data if possible. At minimum, avoid allocating new maps inside the tick.

ui/src/components/three/ThreeRenderer.ts
```diff
@@ -0,0 +1,3905 @@
...
+      const o = i * 6;
+      if (s === undefined || t === undefined) {
+        // Degenerate (shouldn't happen — edges are filtered at build time).
+        ep[o] = ep[o + 1] = ep[o + 2] = 0;
+        ep[o + 3] = ep[o + 4] = ep[o + 5] = 0;
+        continue;
+      }
+      ep[o] = pos[s * 3];
+      ep[o + 1] = pos[s * 3 + 1];
+      ep[o + 2] = pos[s * 3 + 2];
```
ghost, line 1757: ⚠️ `updateEdgePositions` and `updateEdgeAlpha` perform O(Edges) work on the main thread, often multiple times per second (on layout updates). Using `this.nodeIdToIndex.get(e.sourceId)` for every edge adds significant overhead.

We should store the source and target node indices directly on the `ThreeEdge` objects when they are created in `buildEdges` to avoid these map lookups in the hot loops.

ui/src/components/workers/forceLayout3dWorker.ts
ghost, line 382: 🚨 This force is computed every tick. Allocating 4 new `Map` objects and performing multiple map lookups per node (O(Nodes)) will be extremely slow for 100k nodes. 

Pre-allocate typed arrays for the centroids and map node IDs to integer indices once to keep this loop tight and allocation-free.

ui/src/components/three/ThreeRenderer.ts
ghost, line 1757: ⚠️ This loop runs on every layout update (up to 15fps) and performs a Map lookup for every edge. 

We should store the `sourceIdx` and `targetIdx` directly on the `ThreeEdge` objects in `buildEdges` to avoid the overhead of `nodeIdToIndex.get()` in this hot loop.

ui/src/components/three/ThreeRenderer.ts
```diff
@@ -0,0 +1,3905 @@
...
+        ep[o + 5] = pos[tt * 3 + 2];
+        ea[k2] = ea[k2 + 1] = alpha;
+      }
+      (
+        this.edgeGeometry.getAttribute('position') as BufferAttribute
+      ).needsUpdate = true;
+      (
+        this.edgeGeometry.getAttribute('aAlpha') as BufferAttribute
+      ).needsUpdate = true;
+    }
```
ghost, line 5057: ⚠️ This animation runs at 60fps and performs O(Nodes + Edges) work on the main thread, including Map lookups. On a large graph (100k nodes, 300k edges), this loop will take longer than 16ms and hang the UI.

Consider driving the reveal/fly-out logic in the vertex shader by passing a `birthTime` attribute to the GPU.

ghost (commented):
## 🔧 Changes Requested

This is an impressive body of work. Transitioning to Three.js with shader-based rendering is the right architectural choice for handling 100k+ nodes, and the new features like view presets and progressive loading significantly improve the tool's utility.

However, I've flagged a few critical performance issues in the layout worker and the assembly animation that will prevent the graph from scaling to its intended capacity without hanging the browser. Specifically, the per-tick allocations in the community force and the O(N) main-thread work during animations need to be addressed to maintain a smooth 60fps experience on large datasets.

See the inline comments for specific details on the worker and renderer optimizations.

One architectural question: given the move to `d3-force-3d`, did you evaluate using a GPGPU-based layout for the 100k+ regime, or is the current CPU-worker approach meeting the performance targets once the allocation overhead is removed?

<details>

1. **LadybugDB instantiation**: In `ui/src/store/ladybugStore.ts:590`, the database is initialized with `:memory:`. If we expect users to index very large repos, have we checked the peak memory footprint for the WASM heap?
2. **Community Hub heuristic**: In `ui/src/components/colors/communityColors.ts:191`, the $O(N^2)$ substring search for hub nodes is currently running on the main thread. It's probably fine for typical clusters, but Louvain can sometimes produce one giant component that might cause a multi-second stutter during the naming phase.
3. **Embedder Shim**: The `embedderShim` in `storeWorker.ts` is a clean way to keep the ONNX model in one place. The extra postMessage hop is a fair trade-off for avoiding binary duplication.
</details>

ghost:
## 🔄 Review Update
> 31 new commits: f76702d, 7ad4159, a561cf4, ae21473, 8efb928, 83de9d7, f604d8d, 324fd21, 51ebad8, 52d5e1e, 2b7ac0d, eb86805, a49e898, 2ec9187, 143202b, 90891ca, 7c55ab2, 4ac2d80, 0108b81, 75bbd45, c82e3e8, c4e17bc, 073aaf8, a71c6d0, 2d4b9bf, 95beeca, 0abccc2, b5fb0a7, fa6691c, 294184f, b371cf3

Nice progress on finishing the Three.js transition and cleaning up the architecture — moving the Ladybug engine back to main is a good call for Safari support. The new 3D tree layout and the "burst" build animation both look great.

I've resolved the thread on traversal animation as that now correctly uses node indices instead of ID lookups. However, the Map lookups in `updateEdgePositions` and the Map allocations inside the community force tick remain. These will be the primary blockers for the 100k-node target, as they'll likely cause the UI to stutter or the layout to crawl once you hit those scales.

Since the build animation is also performing those same Map lookups every frame, it might be worth tackling the `ThreeEdge` indexing across both features at once.

PR is in a good spot feature-wise, just need to tighten up those hot paths to hit the performance goals.

ui/src/components/three/ThreeRenderer.ts
jatinot: Fixed in 39ce604 — `ThreeEdge` now stores `sourceIdx`/`targetIdx`, resolved once in `buildEdges`. `updateEdgePositions`, `updateEdgeAlpha` (and the edge-cull / edge-pick loops) now read those indices directly instead of `nodeIdToIndex.get()` per edge.

jatinot: Fixed in 39ce604 — `ThreeEdge` now stores `sourceIdx`/`targetIdx`, resolved once in `buildEdges`. `updateEdgePositions`, `updateEdgeAlpha` (and the edge-cull / edge-pick loops) now read those indices directly instead of `nodeIdToIndex.get()` per edge.

ui/src/components/workers/forceLayout3dWorker.ts
jatinot: Fixed in 39ce604 — `makeCommunityForce` now precomputes a dense community-index map (`Int32Array`) plus member counts and reusable centroid accumulators (`Float64Array`) once at force creation (the force is recreated via `buildSimulation` whenever nodes/communities change, so it stays valid). The per-tick loop is now allocation-free with no Map ops — just `.fill(0)` + integer-indexed array reads.

jatinot: Fixed in 39ce604 — `makeCommunityForce` now precomputes a dense community-index map (`Int32Array`) plus member counts and reusable centroid accumulators (`Float64Array`) once at force creation (the force is recreated via `buildSimulation` whenever nodes/communities change, so it stays valid). The per-tick loop is now allocation-free with no Map ops — just `.fill(0)` + integer-indexed array reads.

ui/src/components/three/ThreeRenderer.ts
jatinot: Partly addressed in 39ce604 — the build animation's per-frame edge loop no longer does `nodeIdToIndex.get()` per edge (it reads the stored `sourceIdx`/`targetIdx`), and the BFS adjacency build is lookup-free too. Re: the full vertex-shader `birthTime` rewrite — deferring that: the build replay is a one-time, transient animation triggered after indexing (not a steady-state cost), and on large graphs LOD already engages (>30k). Logged as a follow-up if it proves to be a problem on very large graphs.

ui/src/components/three/ThreeRenderer.ts
jatinot: Partly addressed in 39ce604 — the per-edge `nodeIdToIndex.get()` lookups are gone (ambient rebuilds edges via `updateEdgePositions`, which now uses the stored `sourceIdx`/`targetIdx`). Ambient motion is opt-in (panel toggle) and its amplitude/cost is already eased down for >8k-node graphs. A full GPU/vertex-shader drift is the right long-term answer for 100k+ and is noted as a follow-up; for now the per-edge work is allocation- and lookup-free.

ui/src/components/workers/forceLayout3dWorker.ts
ghost: Nice work on these optimizations. The new `makeCommunityForce` in 39ce6044 precomputes the community mappings and uses pre-allocated typed arrays for the centroids, which completely removes the Map allocations and lookups from the hot tick loop. This should keep the 3D layout smooth even at the 100k node target. I'll mark this as resolved.

ui/src/components/three/ThreeRenderer.ts
ghost: This is much better. By storing `sourceIdx` and `targetIdx` directly on the `ThreeEdge` objects in `buildEdges` (around line 1240 of 39ce6044), you've eliminated the `nodeIdToIndex` lookups in `updateEdgePositions` and `updateEdgeAlpha`. Resolving this one.

ui/src/components/workers/forceLayout3dWorker.ts
ghost: Resolved this one alongside the other community force thread — 39ce6044 introduces the pre-allocated centroids and integer-based indexing in the worker tick.

ui/src/components/three/ThreeRenderer.ts
ghost: Closing this out as well, since 39ce6044 now uses the pre-resolved node indices for edge updates.

ui/src/components/three/ThreeRenderer.ts
ghost: While the build animation still runs on the CPU in 39ce6044, the move to pre-resolved node indices and the addition of `sparkStride` sampling for big graphs significantly reduces the main-thread overhead. The `nodeIdToIndex` Map lookups are gone from the per-edge loops in `updateBuildAnim` (and `updateTraversalAnim` which I previously noted). Looks good, closing this out.

ui/src/components/three/ThreeRenderer.ts
ghost: I'll mark this as resolved since the most expensive part of the animation loop (the Map lookups) has been eliminated in 39ce6044 by using the pre-resolved node indices. The performance should now be acceptable even for larger graphs without needing to move the full fly-out logic to the GPU.

ghost:
## ✅ Review Update — Approved
> 1 new commit: 39ce604

Nice work on the performance optimizations. You've cleared the main blockers by moving the community force to an allocation-free loop and eliminating those Map lookups in the edge rendering path. This looks much leaner now and should handle the 100k-node target without issues.

I've resolved all the outstanding performance threads. With these hot paths tightened up, the Three.js renderer is in a great spot. Ready to merge from my side.

ghost:
## ✅ Review Update — Approved
> 2 new commits: 6444dae, f1b17ad

The latest changes to the `tsconfig` files correctly include the `d3-force-3d` type definitions for the library build. Since the core performance optimizations—specifically the transition to pre-resolved node indices in the renderer's hot loops and the allocation-free community force—are already in place, the Three.js engine is solid.

Everything looks addressed and the PR is ready to merge.

ghost:
## ✅ Review Update — Approved
> 1 new commit: dfc23b0

The transition to a continuous "live build" experience is a massive improvement. Watching the graph assemble itself while indexing—especially with the BFS-staggered fly-out and the new compact `LiveIndexingPanel`—makes the whole process feel much more responsive and interactive.

The architectural shift to off-thread parallel parsing via the `ExtractPool` is exactly what was needed to keep the UI at 60fps during heavy background work. I also really like the new "Onion" preset; the Fibonacci-distributed concentric shells are a great way to visualize the containment hierarchy in 3D.

Performance on the hot paths looks solid—you've maintained the optimizations from the earlier rounds (no Map lookups in `updateEdgePositions`, allocation-free community force) and integrated them cleanly into the live-build logic.

Everything looks addressed and the renderer is in a great spot. Ready to go from my side.

ghost:
## ✅ Review Update — Approved
> 1 new commit: c4ef875

These latest commits bring a lot of great UX polish to the renderer. The proximity-based damping for ambient motion is a clever fix—it was definitely tricky to click nodes while they were drifting, and they settle very naturally now.

I also like the addition of the hover growth in the shader and the lazy-loaded summary card; it makes the graph feel much more interactive. The shift to a sunflower spiral for the 2D Onion layout is also a big readability win over the old projection.

Touch support (pinch-zoom/pan) looks solid, and the refined label gating ensures the overview stays clean even when cranking up the node sizes. 

Everything looks addressed and the performance optimizations from the previous rounds have been preserved. This is in great shape and ready to go.

ghost:
## ✅ Review Update — Approved
> 1 new commit: cdc6f28

The latest changes bring a lot of robustness and interaction polish to the new renderer. 

The worker crash resilience in `ExtractPool` and the `writeChain` in `LadybugGraphStore` are great additions for data integrity—it's much safer now. I also like the explicit WebGL2 validation; failing loudly with a helpful message is much better than a silent blank screen.

The interaction fixes are a nice touch too. The new 3D rotation pivot feels much more stable, and the improvements to label density ensure the overview stays clean even on massive graphs.

Everything looks addressed and the performance/stability optimizations from the earlier rounds have been preserved. This is in great shape for merge.

ghost:
## ✅ Review Update — Approved
> 6 new commits: 95beeca, 0abccc2, b5fb0a7, fa6691c, 294184f, 16d62e1

The latest updates really solidify the robustness of the new renderer. 

The additions to the worker pool—specifically the crash resilience in `ExtractPool` and the `writeChain` in `LadybugGraphStore`—are excellent for data integrity. Handling worker OOMs gracefully and serializing chunked writes prevents the kind of silent failures or data corruption that can be hard to track down in production.

I also like the UX polish throughout:
- The 3D rotation pivot feels much more stable now that it’s anchored to the graph center.
- The proximity-based "freeze" for ambient motion makes node interaction feel very natural.
- The 2D label density caps and the responsive `ViewPresetBar` ensure the UI stays clean regardless of graph size or window dimensions.
- The standalone CSS bundling for detail panels is a great touch for library consumers.

With the WebGL2 validation also in place to prevent silent failures, this is in excellent shape. Performance remains solid on the hot paths we discussed earlier. Ready to go.

ghost:
## ✅ Review Update — Approved
> 1 new commit: 40b691e

The latest fix for vanishing edges in 3D presets is a solid improvement. Moving the depth bias to view-space world units is much more robust than the previous clip-space offset, which was inadvertently shoving edges past the far plane in presets like Planet.

With this stability fix in place alongside the earlier performance optimizations, the renderer is in great shape and ready to go.

ghost:
## ✅ Review Update — Approved
> 1 new commit: 6ad35cf

The latest commits bring a final layer of high-end polish to the Three.js renderer, particularly around the chat traversal UX and rendering quality.

The architectural refinements to the traversal logic are excellent. The new "chain-lightning" pacing for large result sets and the catch-up burst for backlogs ensure the animation stays responsive and never falls behind the agent. I also really like the addition of the partial-trail drawing and the way pending edges stay hidden until crossed—seeing the connections visibly "draw" themselves makes the walk feel much more intentional.

On the rendering side, the two-pass halo fix for glowing nodes is a great catch. Splitting the node material into a solid core (with depth write) and a soft halo pass (without) completely solves the "dark ring" artifact over edges. The transition to absolute alpha encoding for hot edges is also a clever way to keep paths visible on presets like Planet or Onion without needing to raise the global edge opacity multiplier.

Finally, ensuring the layout reseeds and auto-fits after a live build guarantees that the final graph converges to its canonical preset shape regardless of node arrival order.

Everything looks fantastic and the renderer is in a very solid spot. Ready to go.

ghost:
## ✅ Review Update — Approved
> 1 new commit: eea0646

The latest commits bring a final layer of robustness and performance tuning to the Three.js renderer, specifically addressing edge cases in large-scale live builds and memory management.

Nice work on the OOM resilience—explicitly detecting fatal WASM memory errors and halting follow-up queries is a great safeguard to keep the tab stable. The move to lean property streaming (carrying just the `subtype` fields for filters) is also a smart way to keep the JS heap light on massive repos.

A few other highlights:
- The ambient motion cap (20k nodes) and traversal repaint batching are solid wins for maintaining 60fps on large datasets.
- The fix for relational edges in the layout worker ensures the final graph shape is accurate even when links arrive without new nodes.
- Great UX polish on the preset highlights and preserving filter state during search refinements.

This is in excellent shape and ready to go.

dapywell (approved):

PR #404, opened 2026-06-29
feat(export): Expose graph setting defaults

callumyuille:
## Expose graph setting defaults and fix hook warning
🐛 **Bug Fix** · ♻️ **Refactor** · ✨ **Improvement**

This PR moves the graph setting defaults into a standalone, dependency-free module and exposes them through a lightweight utility entry point.

It also fixes a `react-hooks/exhaustive-deps` warning in the Pixi renderer by removing an unused variable from a dependency array.

### Complexity
🟢 Low · `5 files changed, 71 insertions(+), 26 deletions(-)`

This PR is a straightforward structural refactor and cleanup. It moves existing constants to a more accessible module and removes an unused variable from a hook's dependency array with no impact on runtime logic.


ghost (commented):
## ✅ Approved

The changes look solid. Decoupling `GRAPH_SETTING_DEFAULTS` from the hook layer is a clean way to support prop-driven consumers while keeping the library's entry points lightweight.

The removal of `toggleSim` from `PixiGraphCanvas` is safe and correct; it was internal to the layout hook and wasn't exposed via the public `GraphCanvasHandle` or used within the canvas component itself.

<details>

1. **Type duplication**: `ui/src/config/graphSettingDefaults.ts:39`
   I noticed `layoutMode` uses a local type cast. Since `LayoutMode` is also defined in `pixiLayoutWorker.ts` and a few other places, it might be worth centralising that string union in a future cleanup, but it doesn't impact correctness here.

</details>

ghost:
## ✅ Review Update — Approved
> 1 new commit: bf76d79

The updated JSDoc and the addition of the contract test in `bf76d791` address the concern about the diverging defaults perfectly. Framing `GRAPH_SETTING_DEFAULTS` as the viewer's opinionated preset rather than literal component fallbacks makes the intent clear, and the test ensures we won't accidentally drift in the future.

This looks solid and ready to land.

ghost:
## ✅ Review Update — Approved
> 1 new commit: 10bb651

Nice follow-through on the `GraphPlayground` seeding. Using the shared defaults there effectively eliminates the hardcoded drift we were talking about and serves as a good example for how the package is meant to be used.

Everything looks complete and ready to merge.

AnysSido (approved):

PR #405, opened 2026-07-01
fix(ui): drop wayfinder label when its community is hidden

callumyuille:
## Hide wayfinder labels for hidden communities
🐛 **Bug Fix** · ✨ **Improvement** · 🔧 **Chore**

This PR ensures community wayfinder labels are hidden when their constituent nodes are filtered out. 

It introduces a visibility mask that respects previous overlap-cull decisions. This prevents labels from jumping or "churning" (where a neighbor might steal a label slot) when a community is temporarily hidden and then restored.

### Complexity
🟢 Low · `1 file changed, 39 insertions(+), 1 deletion(-)`

Single-file change that fixes a UI bug by adding a targeted visibility pass. It maintains existing architectural constraints regarding label culling and performance.

### Review focus

- **Visibility logic** — verify that labels require both the cull decision and at least one visible node to show.
- **Performance** — confirm the O(N) pass over nodes in `setNodeVisibility` is acceptable for the project's target graph sizes.

### DIFF
--- a/ui/src/components/pixi/PixiRenderer.ts
+++ b/ui/src/components/pixi/PixiRenderer.ts
@@ -1943,9 +1943,19 @@ export class PixiRenderer {
    *  `currentlyShownCommunities`. Hiding / un-hiding a community must
    *  not disturb that decision — it should only reveal or hide the
    *  already-decided labels — so a label shows iff its community was
-   *  chosen by the cull AND still has at least one visible node. This
-   *  keeps hide/un-hide fully symmetric and never evicts an unrelated
-   *  community's label. */
+   *  chosen by the cull AND still has at least one visible node. As long
+   *  as the cached decision holds, hide/un-hide is symmetric and never
+   *  evicts an unrelated community's label.
+   *
+   *  Caveat: a real cull *can* re-run between a hide and its un-hide —
+   *  e.g. a layout-mode toggle (`setLayoutMode`) or a community-data
+   *  rebuild resets `communityVisibilityFrozen`. Because a hidden
+   *  community contributes no centroid, that cull drops it from
+   *  `currentlyShownCommunities` and a neighbour may take its slot, so
+   *  its label won't return on un-hide until the next cull re-decides.
+   *  That's intended — those events are meant to re-pick the label set —
+   *  it just means the symmetry guarantee only spans a hide/un-hide pair
+   *  with no intervening cull. */
   private applyCommunityLabelVisibility(): void {
     if (this.communityLabels.size === 0) return;
     if (!this.communityAssignments) return;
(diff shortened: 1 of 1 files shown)


ghost (commented):
## ✅ Approved

This is a clean fix for the ghost wayfinder label issue. The decision to mask the visibility against the current node set rather than re-running the overlap cull is the right choice—it prevents the "jumping" behavior where labels would shift or swap simply because a neighbor was hidden.

The logic in `applyCommunityLabelVisibility` correctly handles the symmetry between hiding and un-hiding communities while maintaining the visual stability of the labels during transitions.

<details>

1. **Performance**: `ui/src/components/pixi/PixiRenderer.ts:1955`
   The $O(N)$ scan over all nodes is perfectly acceptable here since `setNodeVisibility` already performs a similar pass and this only triggers on actual visibility changes. If we ever needed to optimize this for extremely large graphs, we could pass the `visibleIds` Set directly to `applyCommunityLabelVisibility` to avoid iterating over hidden nodes.

2. **Documentation**:
   The comments explaining the rationale for not re-packing are helpful and clarify the intent for future maintainers.

</details>

ghost:
## ✅ Review Update — Approved
> 1 new commit: 9fc7bd3

Nice work on the follow-up docs. The updated docstring for `applyCommunityLabelVisibility` provides a much clearer picture of the symmetry guarantees. Specifying that the hide/un-hide symmetry is scoped to the current cull decision makes the "fully symmetric" claim more accurate and helps explain why labels might be re-evaluated after a layout-mode toggle or data rebuild.

With that clarification in place, this looks solid and ready to land.

ghost:
## ✅ Review Update — Approved
> 1 new commit: ad2de4e

The latest commit addresses a great catch regarding label scaling. Calling `applyCommunityLabelCounterScale` after updating visibility ensures that wayfinder labels immediately match the current zoom level when they're un-hidden, rather than waiting for the next interaction to trigger a refresh.

This looks solid and ready to land.

ghost:
## ✅ Review Update — Approved
> 1 new commit: 94fccfd

The latest optimization in `applyCommunityLabelVisibility` is a great addition. Skipping the O(N) node loop when the community-label layer is toggled off avoids unnecessary work, especially since `setShowCommunityLabels` handles the refresh when the layer is turned back on.

With the documentation updates, scaling fix, and this performance tweak, this is in excellent shape. Happy to see this land.

ghost:
## ✅ Review Update — Approved
> 2 new commits: 2766e48, c83afb4

Nice catch on the toggle-on synchronization. Ensuring `applyCommunityLabelVisibility` runs when the layer is re-enabled correctly handles communities that were hidden while the layer was off. The refactor to drop the redundant scaling call in `setShowCommunityLabels` is a nice cleanup too, since the visibility pass already handles the refresh.

Everything looks solid and is ready to land.

jatinot (approved):

PR #408, opened 2026-07-06
fix(ui): PR review accuracy — SHA-pinned content, per-field truncation, verdict guardrail + model refresh

jatinot:
## Improve PR review accuracy with SHA-pinning and content paging
🐛 **Bug Fix** · ✨ **Improvement** · ♻️ **Refactor** · 🧪 **Tests**

Fixes architectural data gaps and truncation issues that caused the PR review agent to hallucinate or miss changes. It introduces immutable SHA-pinned fetches, line-numbered content paging for large files, and a safety guardrail that prevents unvalidated approvals.

- Fixes a property mismatch that prevented the agent from receiving file content.
- Pins fetches to SHAs and adds support for fork PRs.
- Implements line-numbered paging and per-field truncation for large files.
- Adds a verdict guardrail to downgrade approvals of uninspected files.
- Refreshes LLM provider model lists and UI fallback logic.

### Complexity
🟠 High · `15 files changed, 1013 insertions(+), 176 deletions(-)`

This PR spans the entire PR review lifecycle: from how data is indexed in the graph, to how tools fetch and paginate content, up to the agent's prompt and the final validation of the review verdict. It introduces new foundational patterns like SHA-pinning and line-based content paging that are critical for LLM accuracy and requires careful coordination between the backend graph, the GitHub client, and the agent runner.

### Tests
🧪 Includes extensive new tests for content paging, SHA resolution, fork retries, and the verdict enforcement guardrail.

### Review focus

- **Line-numbering logic** — Verify `packContent` in `prTools.ts` correctly generates 1-based line numbers that align with GitHub's expectations for review comments.
- **Verdict downgrade logic** — Ensure `enforceInspectionCoverage` in `reviewParse.ts` robustly handles different JSON fencing styles (`json:review` vs `json`).
- **Fork PR threading** — Check that `headRepo` is correctly captured by the indexer and utilized by `PRClient` to target the correct repository for fork head commits.

---

<details>

### Data Flow & Persistence
Previously, the indexer wrote `baseSha`/`headSha` to the graph, but the review tools looked for `base_sha`/`head_sha`, leading to a silent fallback to branch names (which shift) or "PR client not available" errors. This PR standardizes on `camelCase` for graph properties while providing a `strProp` helper in the tools to handle legacy or mismatched keys gracefully.

### Truncation Strategy
We moved away from a global `truncate()` on the final JSON string, which often produced unparseable blobs. The new `get_pr_file_change` tool implements per-field budgets:
1. **Diffs** are truncated with a note to use the "new" version for full context.
2. **File Content** is paged using a 1-based line range.
3. Each field carries a `*_truncated` flag, allowing the agent to detect partial data and request the next page using `startLine`.

### Review Integrity
The `enforceInspectionCoverage` helper acts as a late-stage guardrail. During the agent's execution loop, we track every file passed to `get_pr_file_change`. If the agent emits an `APPROVE` verdict but the tracking set doesn't cover all files in `pr.files`, the verdict is programmatically downgraded to `COMMENT` to prevent the agent from "approving" code it never saw.

</details>

### DIFF
--- a/ui/src/chat/prTools.ts
+++ b/ui/src/chat/prTools.ts
@@ -24,12 +24,71 @@ import { z } from 'zod';
 import type { GraphStore } from '../store/types';
 import type { PRClient } from '../pr/client';
 
-const MAX_RESULT_CHARS = 6000;
-const MAX_SOURCE_CHARS = 8000;
+const MAX_RESULT_CHARS = 16_000;
+/** Per-field budgets for get_pr_file_change. Each field is bounded
+ *  independently and flagged when cut — never truncate the combined JSON
+ *  mid-stream, or the model reviews half a payload without knowing it. */
+const MAX_DIFF_CHARS = 12_000;
+const MAX_CONTENT_CHARS = 16_000;
+/** Safety net for the combined get_pr_file_change payload. */
+const MAX_FILE_CHANGE_CHARS = 48_000;
+/** Max compact file entries returned by get_pull_request. */
+const MAX_FILE_ENTRIES = 300;
 
 function truncate(text: string, limit: number): string {
   if (text.length <= limit) return text;
   return text.slice(0, limit) + `\n...[truncated, ${text.length} chars total]`;
+}
+
+/**
+ * Slice file content to a 1-based inclusive line range, numbered cat -n
+ * style so the model can cite exact new-file line numbers, bounded by a
+ * character budget. Reports the range served and where to resume.
+ */
+function packContent(
+  content: string,
+  startLine: number | undefined,
+  endLine: number | undefined,
+  budget: number,
+): {
+  text: string;
+  lines: string;
+  truncated: boolean;
+  next_start_line?: number;
+} {
+  const all = content.split('\n');
+  const total = all.length;
+  const start = Math.min(Math.max(1, startLine ?? 1), total);
+  const end = Math.min(total, endLine && endLine >= start ? endLine : total);
+  const out: string[] = [];
+  let used = 0;
+  let last = start - 1;
+  for (let i = start; i <= end; i++) {
+    const line = `${String(i).padStart(5)}\t${all[i - 1]}`;
+    if (used + line.length + 1 > budget) break;
+    out.push(line);
+    used += line.length + 1;
+    last = i;
+  }
+  const truncated = last < end;
+  return {
+    text: out.join('\n'),
+    lines: `${start}-${last} of ${total}`,
+    truncated,
+    ...(truncated ? { next_start_line: last + 1 } : {}),
+  };
+}
+
+/** Read a string property from a node's properties bag. */
+function strProp(
+  props: Record<string, unknown> | undefined,
+  ...keys: string[]
+): string | undefined {
+  for (const k of keys) {
+    const v = props?.[k];
+    if (typeof v === 'string' && v) return v;
+  }
+  return undefined;
 }
 
 export function makePRTools(store: GraphStore, prClient?: PRClient | null) {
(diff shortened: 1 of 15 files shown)


ghost (commented):
## 🔧 Changes Requested

Overall, these changes significantly raise the bar for the review agent's reliability. Moving to line-numbered snapshots and immutable SHA pinning makes the agent's feedback much more trustworthy, and the fork-PR support is a great addition.

I've requested changes primarily to address a logic error where the `truncate` safety net produces invalid JSON. For a tool like `get_pull_request`, which is expected to return up to 300 files, hitting this limit is a high probability, and returning broken JSON to the LLM will cause it to hallucinate or fail the turn.

See the inline comments for specific suggestions on handling these budgets more gracefully.

### Broader Observations

- **Guardrail Consistency**: The `enforceInspectionCoverage` logic is excellent for the direct "Run Review" action. However, it isn't currently applied when the chat agent delegates to the `code_reviewer` sub-agent tool (in `ui/src/chat/subagents.ts`). It’s worth considering if that tool should also track and enforce this, or if we're okay with the chat path being "lower trust" for now.
- **Recursion Limits**: The `recursionLimit` is set to 80 in `runPRReview`. While this gives the agent plenty of room to page through large files, keep an eye on latency/token costs for very large PRs.

<details>

1. **Patch Truncation**: `ui/src/pr/indexer.ts:37`
   Increasing `MAX_PATCH_CHARS` to 50k is a good move for visibility, but note that this increases the storage footprint for every PR indexed. If storage becomes a concern, we might want to move these to a separate blob table.
2. **Helper Utility**: `ui/src/chat/prTools.ts:83`
   `strProp` is a clean way to handle the various property naming conventions coming out of the indexer.
3. **Safety Net**: `ui/src/chat/prTools.ts:34`
   The `MAX_FILE_CHANGE_CHARS` safety net is quite high (48k). Some providers (like older GPT-4 models) might struggle if the total prompt context gets too bloated, though the 2026-07 models you've added should handle this easily.

</details>

ui/src/chat/prTools.ts
```diff
@@ -59,13 +118,33 @@ export function makePRTools(store: GraphStore, prClient?: PRClient | null) {
...
+            file_count: files.length,
+            ...(omitted
+              ? {
+                  files_omitted: omitted,
+                  note: `Showing first ${MAX_FILE_ENTRIES} of ${files.length} files`,
+                }
+              : {}),
           }),
           MAX_RESULT_CHARS,
         );
```
ghost: 🚨 The `truncate` function at the end of this tool (and others) will return a broken JSON string if it triggers, as it simply slices the stringified JSON and appends a suffix. While the field-level truncation in `get_pr_file_change` helps, `get_pull_request` and `summarize_pr_changes` are likely to hit this limit for large PRs. We should truncate the `files` array until the JSON is within budget instead of truncating the final string.

```suggestion
        const omitted = Math.max(0, files.length - MAX_FILE_ENTRIES);
        let result = {
          pr: node,
          files: files.slice(0, MAX_FILE_ENTRIES),
          file_count: files.length,
          ...(omitted
            ? {
                files_omitted: omitted,
                note: `Showing first ${MAX_FILE_ENTRIES} of ${files.length} files`,
              }
            : {}),
        };

        // Safety: if the JSON is still too large, drop file entries one by one until it fits
        // (better to have fewer valid entries than broken JSON).
        let json = JSON.stringify(result);
        while (json.length > MAX_RESULT_CHARS && result.files.length > 0) {
          result.files.pop();
          result.files_omitted = (result.files_omitted || 0) + 1;
          result.note = `Showing first ${result.files.length} of ${files.length} files`;
          json = JSON.stringify(result);
        }

        return json;
      },
```

ui/src/chat/prTools.ts
```diff
@@ -87,16 +166,29 @@ export function makePRTools(store: GraphStore, prClient?: PRClient | null) {
...
           const fileId = rel.relationship.target_id;
           if (!fileId) continue;
           const incoming = await store.traverse(fileId, 'incoming', 2);
-          blastRadius[fileId] = incoming;
+          blastRadius[fileId] = {
+            dependents: incoming.length,
+            sample: incoming
+              .slice(0, 10)
+              .map((d) => `${d.node.type}:${d.node.name || d.node.id}`),
+          };
```
ghost: ⚠️ The `fileId` (which is often an internal graph ID) is less helpful for the LLM than the file path. Using the path as the key makes the tool output more intuitive for the agent.

```suggestion
          const path = (rel.relationship.properties?.path as string) || fileId;
          const incoming = await store.traverse(fileId, 'incoming', 2);
          blastRadius[path] = {
            dependents: incoming.length,
            sample: incoming
              .slice(0, 10)
              .map((d) => `${d.node.type}:${d.node.name || d.node.id}`),
          };
```

ui/src/chat/prTools.ts
jatinot: Fixed in 9109491 — get_pull_request now drops trailing file entries until the JSON fits MAX_RESULT_CHARS instead of slicing the stringified payload, so the model always receives valid JSON. Applied the same shrink-until-fits approach to summarize_pr_changes. Covered by a new 300-file over-budget test.

ui/src/chat/prTools.ts
jatinot: Fixed in 9109491 — blast_radius is now keyed by the file path from the CHANGES edge properties (falling back to the node ID only when no path is stored). Added a test asserting the path key is used.

ui/src/chat/prTools.ts
ghost: Fixed in 9109491. I've updated the tool to drop trailing file entries until the JSON fits within the character limit, ensuring the model always receives valid JSON. There's also a new test case in `prTools.test.ts` covering the over-budget scenario. Resolving this one.

ui/src/chat/prTools.ts
ghost: Good call. The `blast_radius` is now keyed by the file path (via 9109491), which is much more useful for the agent. I've also applied the same "shrink-until-fits" logic here to prevent broken JSON on large PRs. Closing this out.

ghost:
## ✅ Review Update — Approved
> 1 new commit: 9109491

Nice work addressing the feedback in the latest commit. The new approach to truncation—dropping entries one by one until the payload fits—is much safer and ensures the agent always gets valid JSON. I also appreciate switching the blast radius keys to file paths; that makes the output much more useful for the model.

I've resolved the two outstanding threads and the new test coverage looks solid. This is looking good to go.

PR #420, opened 2026-08-06
feat: add java support

artworx:
I'm not sure if you accept contributions out of the blue. I was missing full Java support, and here is a version that works for my use case. 

## Summary

Adds full Java extraction support to the OpenTrace Python agent, including symbols, method calls, imports, inheritance, and graph relationships.

## Changes

  - Add a Tree-sitter Java extractor.
  - Extract classes, interfaces, enums, methods, constructors, variables, calls, and inheritance.
  - Resolve Java imports against repository source paths.
  - Integrate Java processing into the indexing pipeline and CLI.
  - Add Java extractor, import-resolution, and graph-store tests.
  - Add the tree-sitter-java dependency.
  - Update language-support documentation to distinguish CLI and browser capabilities.

## Test plan

  - make test
  - make lint
  - make license-check

  All checks pass. 

## Checklist

  - [x] Tests pass
  - [x] Documentation updated
  - [x] No breaking changes
  - [x] License headers valid

### DIFF
--- a/agent/src/opentrace_agent/cli/main.py
+++ b/agent/src/opentrace_agent/cli/main.py
@@ -265,10 +265,29 @@ def _clean_stale_staging(staging_db: str) -> None:
 
 
 def _seed_staging_from_live(db_path: str, staging_db: str) -> None:
-    """Clone live DB + WAL into staging so a new index appends rather than replaces."""
+    """Clone live DB + WAL into staging so a new index appends rather than replaces.
+
+    Forces a WAL checkpoint on the live DB first so the copy captures a
+    self-contained snapshot.  Without this, a large WAL accumulated from
+    prior writes would be duplicated into the staging file, doubling the
+    on-disk footprint and seeding the staging DB with dead pages that
+    LadybugDB never reclaims.
+    """
     live = Path(db_path)
     if not live.exists():
         return
+
+    # Checkpoint the live DB so the WAL is folded into the main file
+    # before we byte-copy it.  Opened read_only=False briefly just for
+    # the CHECKPOINT command; the exclusive index lock is already held.
+    from opentrace_agent.store import GraphStore
+
+    try:
+        with GraphStore(db_path) as gs:
+            gs.checkpoint()
+    except Exception as exc:
+        # Non-fatal: the copy still works, just with a potentially large WAL.
+        click.echo(f"Note: pre-copy checkpoint skipped ({exc})", err=True)
 
     shutil.copy2(live, staging_db)
     live_wal = Path(db_path + ".wal")
@@ -353,6 +372,148 @@ def _swap_staging_into_place(staging_db: str, db_path: str) -> None:
         _safe_unlink(live_wal, context="post-rename cleanup")
 
 
+def _do_index(
+    *,
+    staging_db: str,
+    source_path: Path,
+    repo_id: str,
+    batch_size: int,
+    verbose: bool,
+    extra_metadata: dict[str, object] | None,
+    on_event: "Callable[[object], None] | None",
+    GraphStore,
+    GraphStoreAdapter,
+    PipelineInput,
+    run_pipeline,
+) -> float:
+    """Core indexing logic: open staging, run pipeline, flush, checkpoint."""
+    with GraphStore(staging_db) as graph_store:
+        store = GraphStoreAdapter(graph_store, batch_size=batch_size)
+
+        click.echo(f"Indexing {source_path} ...")
+        t0 = time.monotonic()
+
+        inp = PipelineInput(path=str(source_path), repo_id=repo_id)
+
+        last_result = None
+        for event in run_pipeline(inp, store=store):
+            _print_event(event, verbose)
+            if getattr(event, "result", None) is not None:
+                last_result = event.result
+            if on_event is not None:
+                try:
+                    on_event(event)
+                except Exception:
+                    logging.getLogger(__name__).debug("on_event callback raised; ignoring", exc_info=True)
+
+        elapsed = time.monotonic() - t0
+
+        metadata = _collect_metadata(source_path, repo_id, elapsed, last_result)
+        if extra_metadata:
+            metadata.update(extra_metadata)
+        graph_store.save_metadata(metadata)
+
+        store.flush()
+
+        # Checkpoint the staging DB before the atomic swap so the
+        # WAL is folded into the main file.  This keeps the swapped-
+        # in DB compact and avoids carrying a bloated WAL sidecar
+        # that would be byte-copied again on the *next* index run.
+        graph_store.checkpoint()
+
+    return elapsed
+
+
+def _is_pk_collision(exc: Exception) -> bool:
+    """Return True if *exc* is a LadybugDB duplicate-primary-key error."""
+    msg = str(exc).lower()
+    return "primary key" in msg and ("duplicat" in msg or "uniqueness constraint" in msg)
+
+
+def _index_into_staging(
+    *,
+    staging_db: str,
+    db_path: str,
+    source_path: Path,
+    repo_id: str,
+    batch_size: int,
+    verbose: bool,
+    extra_metadata: dict[str, object] | None,
+    on_event: "Callable[[object], None] | None",
+    GraphStore,
+    GraphStoreAdapter,
+    PipelineInput,
+    run_pipeline,
+) -> float:
+    """Seed staging from live, delete old rows, index, checkpoint.
+
+    If the import hits a LadybugDB primary-key collision (stale internal
+    row-ID metadata baked into the DB file), discard the corrupted staging
+    copy and retry into a **fresh empty** staging DB.  The live DB is left
+    untouched so other repos' data is preserved; they will be re-seeded
+    from the live DB on their next individual index run.
+    """
+    _seed_staging_from_live(db_path, staging_db)
+
+    # Wipe this repo's rows so the bulk-import path starts clean.
+    with GraphStore(staging_db) as graph_store:
+        deleted = graph_store.delete_repo(repo_id)
+        # Flush the WAL after the bulk delete so that freed pages
+        # are consolidated before the connection closes.  This
+        # also prevents the subsequent fresh connection from
+        # seeing stale internal primary-key metadata.
+        graph_store.checkpoint()
+    if deleted.get("nodes_deleted"):
+        click.echo(f"Re-index: cleared {deleted['nodes_deleted']} existing nodes for '{repo_id}'")
+
+    try:
+        return _do_index(
+            staging_db=staging_db,
+            source_path=source_path,
+            repo_id=repo_id,
+            batch_size=batch_size,
+            verbose=verbose,
+            extra_metadata=extra_metadata,
+            on_event=on_event,
+            GraphStore=GraphStore,
+            GraphStoreAdapter=GraphStoreAdapter,
+            PipelineInput=PipelineInput,
+            run_pipeline=run_pipeline,
+        )
+    except RuntimeError as exc:
+        if not _is_pk_collision(exc):
+            raise
+
+    # ------------------------------------------------------------------
+    # Retry: the live DB has corrupted internal primary-key state that
+    # survived the delete_repo + checkpoint + fresh-connection sequence.
+    # Discard the tainted staging copy and re-index into a blank DB.
+    # Other repos will be absent from this staging file, but the live DB
+    # is untouched — subsequent index runs for those repos will re-seed
+    # their data from the live copy.
+    # ------------------------------------------------------------------
+    click.echo(
+        "Primary-key collision detected in seeded staging DB; retrying with a fresh empty database ...",
+        err=True,
+    )
+    _safe_unlink(staging_db, context="pk-collision retry")
+    _safe_unlink(staging_db + ".wal", context="pk-collision retry")
+
+    return _do_index(
+        staging_db=staging_db,
+        source_path=source_path,
+        repo_id=repo_id,
+        batch_size=batch_size,
+        verbose=verbose,
+        extra_metadata=extra_metadata,
+        on_event=on_event,
+        GraphStore=GraphStore,
+        GraphStoreAdapter=GraphStoreAdapter,
+        PipelineInput=PipelineInput,
+        run_pipeline=run_pipeline,
+    )
+
+
 def _run_indexing_pipeline(
     *,
     source_path: Path,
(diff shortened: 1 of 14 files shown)


agent/src/opentrace_agent/cli/main.py
```diff
@@ -353,6 +372,148 @@ def _swap_staging_into_place(staging_db: str, db_path: str) -> None:
...
+    # their data from the live copy.
+    # ------------------------------------------------------------------
+    click.echo(
+        "Primary-key collision detected in seeded staging DB; retrying with a fresh empty database ...",
+        err=True,
+    )
+    _safe_unlink(staging_db, context="pk-collision retry")
+    _safe_unlink(staging_db + ".wal", context="pk-collision retry")
+
+    return _do_index(
```
callumyuille, line 502: **[Critical — blocking] PK-collision retry silently destroys other repos' data.**

On a collision this branch unlinks the seeded staging DB and re-runs `_do_index` against the now-nonexistent path, so `GraphStore(staging_db)` creates a **fresh, empty** DB containing only the repo being indexed. But `_run_indexing_pipeline` still unconditionally swaps staging over live via `_swap_staging_into_place` → `os.replace` (line 576) — there's no flag distinguishing "seeded staging" from "fresh-retry staging".

Net effect: if repos A and B are already indexed and indexing repo C hits a collision, the live DB is atomically replaced with a DB containing only C. A and B are gone, and the command exits 0.

The comments here and in the `_index_into_staging` docstring ("the live DB is left untouched so other repos' data is preserved; they will be re-seeded on their next index run") are unfortunately wrong in the way that matters: live is untouched only *until* the swap, and after it there's nothing left to re-seed from.

A retry that starts from an empty DB needs to either re-seed from live again (minus the colliding rows) or skip the swap and surface an error instead.

agent/src/opentrace_agent/sources/code/import_analyzer.py
```diff
@@ -593,6 +593,134 @@ def _parse_ts_reexport(
...
+        # Check if any known file ends with this suffix
+        for known in known_files:
+            if known.endswith("/" + suffix) or known == suffix:
+                return known
+        idx = pos + 1
+    # Last resort: match just the filename
+    filename = candidate.rsplit("/", 1)[-1]
+    for known in known_files:
+        if known.endswith("/" + filename):
+            return known
```
callumyuille, line 720: **[Critical — blocking] Basename-only matching resolves external imports to unrelated local files.**

The suffix loop degrades all the way down to the bare filename (`known.endswith("/" + suffix)` with `suffix == "Test.java"`), and the explicit last-resort below does a pure basename match. There's no validation of the package segments against the file's directory path and no external-prefix check.

So `import org.junit.Test` resolves as *internal* to any local `Test.java`, `java.util.List` to any `List.java`, etc. — and since `_parse_java_import` only emits a Dependency node when resolution fails, these common imports produce zero Dependency nodes and a bogus internal IMPORTS edge instead. This fires constantly on common class names (`Test`, `List`, `Service`, `Builder`...). Same issue in the static-import branch.

Notably the Go resolver's docstring in this same file (lines 272–275) warns against exactly this: "Matching only the last segment would wrongly link unrelated third-party packages."

Fix direction: require the matched file's path suffix to cover the full package path (e.g. `org/junit/Test.java`), and treat well-known external prefixes (`java.`, `javax.`) as never-internal. The existing tests don't cover the collision case — worth adding one with a local `List.java` alongside `import java.util.List`.

agent/src/opentrace_agent/sources/code/import_analyzer.py
```diff
@@ -593,6 +593,134 @@ def _parse_ts_reexport(
...
+    idx = 0
+    while True:
+        pos = candidate.find("/", idx)
+        if pos == -1:
+            break
+        suffix = candidate[pos + 1 :]
+        # Check if any known file ends with this suffix
+        for known in known_files:
+            if known.endswith("/" + suffix) or known == suffix:
+                return known
```
callumyuille, line 714: **[Critical — blocking] Per-import full scan of `known_files` is O(imports × files).**

`_resolve_java_path` iterates the entire `known_files` set inside the suffix loop (and again in the basename fallback), and the wildcard branch does another full scan. This is the exact quadratic blow-up that `build_go_dir_index` was introduced to prevent — see the comment at `pipeline/processing.py:102-105` ("~5e8 iterations (minutes) on Grafana-scale repos"). Go builds its index once per repo and does O(depth) dict lookups; the Java path gets only the raw set (`processing.py:306`).

Suggest building an equivalent one-time index for Java (e.g. class-name → paths, or a path-suffix trie) in `processing.py` and passing it in, mirroring the Go approach. That would also make fixing the basename-collision issue above cheaper, since you'd validate package segments against a small candidate list instead of scanning everything.

agent/src/opentrace_agent/sources/code/import_analyzer.py
```diff
@@ -593,6 +593,134 @@ def _parse_ts_reexport(
...
+        # import com.example.service.* → try to find files under that package
+        dir_path = import_path.replace(".", "/")
+        # Try to find any known file under this directory (for the IMPORTS edge)
+        for known in known_files:
+            if known.startswith(dir_path + "/") or ("/" + dir_path + "/") in known:
+                # Map the package name as an alias to the first match
+                alias = import_path.rsplit(".", 1)[-1] if "." in import_path else import_path
+                result.setdefault(alias, known)
+                break
+        return
```
callumyuille, line 664: **Wildcard imports: non-deterministic resolution + external wildcards silently dropped.**

Two issues in this branch:

1. `known_files` is a `set[str]`, so `for known in known_files: ... break` picks a hash-order-dependent first match. Since `PYTHONHASHSEED` isn't pinned (the Go docstring at lines 281–284 says so explicitly), the same repo can produce a *different* IMPORTS edge on different index runs. Sorting the candidates (or resolving against a proper index, per the other comment) makes this deterministic.
2. If nothing matches — the normal case for `import java.util.*` — the function returns without ever touching `external`, so external wildcard imports produce no Dependency node at all.

Also worth noting: the substring test `("/" + dir_path + "/") in known` means `import java.util.*` would falsely resolve *internally* in any repo containing a `src/main/java/util/` directory, since that path contains `/java/util/`.

agent/src/opentrace_agent/cli/main.py
```diff
@@ -265,11 +265,30 @@ def _clean_stale_staging(staging_db: str) -> None:
...
     if not live.exists():
         return
 
+    # Checkpoint the live DB so the WAL is folded into the main file
+    # before we byte-copy it.  Opened read_only=False briefly just for
+    # the CHECKPOINT command; the exclusive index lock is already held.
+    from opentrace_agent.store import GraphStore
+
+    try:
+        with GraphStore(db_path) as gs:
```
callumyuille, line 286: **Pre-copy checkpoint opens the live DB read-write and runs DDL against it.**

`GraphStore(db_path)` defaults to `read_only=False`, which makes `__init__` run `_ensure_schema()` — `CREATE ... IF NOT EXISTS`, a guarded `ALTER TABLE`, `INSTALL FTS`, and a guarded `CREATE_FTS_INDEX` — against the *live* DB before anything has been staged.

On a healthy, current-schema DB this is effectively a no-op (everything is `IF NOT EXISTS` / try-except guarded). But if the live DB has an older schema (no `search_text` column / FTS index — and the store docs say schema isn't migrated), this open mutates live in place before any swap, which undercuts the stage-then-atomically-swap design the rest of this pipeline is careful about.

If LadybugDB supports it, a `read_only` open for the CHECKPOINT call — or a minimal open path that skips `_ensure_schema()` — would keep the pre-swap phase strictly non-mutating.

agent/src/opentrace_agent/store/graph_store.py
```diff
@@ -338,6 +338,10 @@ def import_batch(
         try:
             return self._import_batch_via_copy(nodes, relationships)
         except Exception as exc:
+            message = str(exc).lower()
+            if "primary key" in message and ("duplicat" in message or "uniqueness constraint" in message):
+                logger.error("Bulk COPY import hit a primary-key collision; refusing unsafe per-row MERGE fallback")
+                raise
```
callumyuille, line 344: **Behavior now contradicts the documented MERGE-fallback contract, and the collision predicate is duplicated.**

Two things here:

1. `import_batch` now hard-raises on PK collisions, but the docstring just above (line 331: "Falls back to per-row MERGE for the batch on any failure (e.g. a primary key collision...)") and the `__init__` invariant comment (lines 175–176: "a false negative merely costs a COPY failure + MERGE-fallback retry") still describe the old behavior. With this change, a false negative in the `_node_ids` mirror is no longer a cheap retry — it propagates up as a collision. Both comments need updating to match.
2. The predicate `"primary key" in msg and ("duplicat" in msg or "uniqueness constraint" in msg)` is duplicated verbatim in `main.py:_is_pk_collision`. If one copy is ever changed without the other, the store raises but main.py doesn't recognize it as a collision and the whole index run fails hard with no retry. Suggest exporting a single `is_pk_collision(exc)` helper from this module and using it in both places.

Related fragility: `main.py:483` catches only `RuntimeError`, but this `raise` re-raises the original exception whatever its type — a collision surfacing as any other exception type already bypasses the retry today.

agent/src/opentrace_agent/sources/code/extractors/java_extractor.py
```diff
@@ -0,0 +1,538 @@
...
+                symbols.append(sym)
+        elif child.type == "interface_declaration":
+            sym = _extract_interface(child)
+            if sym:
+                symbols.append(sym)
+        elif child.type == "enum_declaration":
+            sym = _extract_enum(child)
+            if sym:
+                symbols.append(sym)
+    return symbols
```
callumyuille, line 76: **`record_declaration` and `annotation_type_declaration` are not handled — a Java 17+ record file extracts zero symbols.**

`_walk_program` (and the nested-type dispatch in `_extract_class_body`) handles only `class_declaration`, `interface_declaration`, and `enum_declaration`. I verified empirically: a file containing `public record Point(int x, int y) { public int sum() {...} }` returns an empty symbol list — including the record's methods — so record-heavy modern codebases index as bare File nodes. Nested records inside classes are dropped too, and `@interface` types likewise.

Records at minimum are worth handling before merge given how common they are in Java 17+ code; they can largely share the `class_declaration` path (tree-sitter gives them a `body` with `method_declaration`s the same way).

agent/src/opentrace_agent/sources/code/extractors/java_extractor.py
```diff
@@ -0,0 +1,538 @@
...
+
+
+def _extract_type_signature(params_node: tree_sitter.Node) -> str:
+    """Build a type signature from Java formal parameters, e.g. ``(String,int)``."""
+    types: list[str] = []
+    for child in params_node.children:
+        if child.type == "formal_parameter":
+            type_node = child.child_by_field_name("type")
+            if type_node:
+                types.append(_normalize_java_type(type_node))
```
callumyuille, line 319: **C-style array parameters lose their dimensions, collapsing distinct overloads.**

For `formal_parameter` only the `type` field is read. tree-sitter-java puts C-style brackets in a separate `dimensions` field on the parameter (for `int a[]`: `type=int`, `dimensions="[]"`), so `void h(int a)` and `void h(int a[])` both produce `type_sig=(int)` — verified by running the extractor. Since `type_signature` exists precisely for overload disambiguation, one of the two overloads collides and is silently dropped downstream. (Java-style `int[] a` works fine, since there the dims are part of the type node.)

Fix: after `_normalize_java_type(type_node)`, append the text of `child.child_by_field_name("dimensions")` if present.

agent/src/opentrace_agent/sources/code/extractors/java_extractor.py
```diff
@@ -0,0 +1,538 @@
...
+            if current.end_point.row != expected_row:
+                break
+            text = current.text.decode()
+            if text.startswith("//"):
+                lines.append(text[2:].strip())
+            expected_row = current.start_point.row - 1
+            current = current.prev_named_sibling
+        if lines:
+            lines.reverse()
+            return "\n".join(lines)
```
callumyuille, line 465: **A trailing `//` comment on the preceding member gets attached as the next method's docs.**

The line-comment scan checks only row adjacency (`current.end_point.row != expected_row → break`), never whether the comment starts its own line. So:

```java
void first() {
    ...
} // legacy hack, do not touch
void second() {}
```

gives `second.docs == "legacy hack, do not touch"` (verified empirically). Adding a check that the comment's `start_point.row` is on a different row than the previous member's `end_point.row` (i.e. the comment owns its line) fixes it.

agent/src/opentrace_agent/sources/code/extractors/java_extractor.py
```diff
@@ -0,0 +1,538 @@
...
+                children.append(sym)
+        elif child.type == "interface_declaration":
+            sym = _extract_interface(child)
+            if sym:
+                children.append(sym)
+        elif child.type == "enum_declaration":
+            sym = _extract_enum(child)
+            if sym:
+                children.append(sym)
+    return children
```
callumyuille, line 242: **No fields, interface constants, or enum constants are extracted — Java yields zero Variable nodes.**

`_extract_class_body` handles methods, constructors, and nested types, but `field_declaration`, `constant_declaration`, and `enum_constant` nodes are skipped, and `symbol.variables` is never populated anywhere in this extractor. The pipeline does consume `symbol.variables` into Variable nodes (`processing.py:388-389` for class fields), so the data would flow if extracted.

To be fair, this matches Go and TypeScript (only the Python extractor populates `variables` today), so I'd call it a parity gap rather than a blocker — fine to land as a follow-up if you'd prefer, but worth a TODO or a line in `docs/reference/languages.md` noting the limitation either way.

agent/src/opentrace_agent/sources/code/extractors/java_extractor.py
```diff
@@ -0,0 +1,538 @@
...
+        # check the parent's children list for a block_comment just before us.
+        parent = node.parent
+        if parent is not None:
+            prev = None
+            for child in parent.children:
+                if child.id == node.id:
+                    break
+                if child.is_named:
+                    prev = child
+            sibling = prev
```
callumyuille, line 437: **Nit: this fallback is a no-op and can be deleted.**

The loop over `parent.children` keeping the last named child before `node` computes exactly what `prev_named_sibling` already returned — same parent, same child list, same named filter (block comments *are* named nodes in tree-sitter-java, contrary to the comment above about unnamed siblings). I verified it executes for the first member of a class body but always reassigns `None` to `None`; when a `/** doc */` is present, `prev_named_sibling` already finds it directly.

agent/src/opentrace_agent/sources/code/extractors/java_extractor.py
```diff
@@ -0,0 +1,538 @@
...
+    Captures bare calls (``foo()``), dotted calls (``obj.foo()``),
+    and chained calls (``a.b().c()``).
+    """
+    calls: list[CallRef] = []
+    own = _extract_call_from_node(node)
+    if own is not None:
+        calls.append(own)
+    for child in node.children:
+        calls.extend(_collect_calls(child))
+    return calls
```
callumyuille, line 538: **Anonymous-class methods are never emitted as symbols; their calls attach to the enclosing method.**

`_collect_calls` recurses into all children with no boundary at `class_body`/`object_creation_expression`, so in `void start() { new Runnable() { public void run() { foo(); } }.run(); }`, `foo()` becomes a call from `start`. That part is actually consistent with the Python/Go extractors, which deliberately recurse into lambda/closure bodies the same way — so I wouldn't change the attribution.

The Java-specific gap is the other half: unlike lambdas, anonymous classes contain real named `method_declaration`s, and nothing in `_walk_program`/`_extract_class_body` ever looks inside expression trees for class bodies, so `run()` here is never extracted as a symbol at all (method-local named classes are missed for the same reason). For callback-heavy code those methods are invisible in the graph. Fine as a documented follow-up, but worth deciding explicitly rather than by omission.

callumyuille (changes requested):
Hey @artworx thanks for your contribution - Java support is one of the most-requested additions, and this is a  solid foundation: the extractor structure follows the existing per-language pattern nicely, the test coverage on the happy paths is thorough, and things like overload disambiguation via type signatures and the Javadoc handling show real care. We verified each issue below against the branch (several empirically, by running the extractor) before writing this up.

That said, we found a handful of things that need to be addressed before this can merge, plus some smaller items. Inline comments have the details; here's the shape of it:

**Blocking:**
- The PK-collision retry path in `cli/main.py` can silently destroy previously indexed repos' data (staging is recreated empty but still swapped over the live DB). This is the big one.
- `_resolve_java_path` matches on basename alone, so common external imports (`java.util.List`, `org.junit.Test`, ...) resolve to unrelated local files and emit no Dependency nodes.
- Java import resolution does a per-import full scan of `known_paths` — the quadratic pattern `build_go_dir_index` exists to avoid. On large repos this is minutes of pure import resolution.
- Java 17+ `record` declarations extract zero symbols.

**Should fix (correctness/determinism):**
- Wildcard imports resolve non-deterministically across runs, and external wildcards are dropped.
- C-style array params (`int a[]`) lose dimensions and collapse overloads.
- A trailing `//` comment on the previous member becomes the next method's docs.
- `import_batch`'s docstring/invariants still promise the MERGE fallback that the new code removed, and the collision predicate is duplicated between `graph_store.py` and `main.py`.
- The pre-copy checkpoint opens the live DB read-write (DDL runs against live data).

**Non-blocking / fine as follow-ups if you prefer:**
- No fields/enum constants → zero Variable nodes (this matches Go/TS today, so it's a parity gap with Python rather than a regression — a doc note would suffice).
- Anonymous-class methods are never emitted as symbols.
- A small dead-code nit in `_extract_javadoc`.

Happy to discuss approach on any of these — especially the staging/retry one, where the right fix probably deserves a quick design conversation. And if you'd rather we handle the LadybugDB staging-path items (they're a bit deep in pipeline internals that predate this PR), just say the word and we'll take those, keeping the Java extractor work yours. Thanks again for the contribution!

